PR #1, opened 2015-05-26
Feature/more list functions

hallettj:
I added a new function for Lists in each of four commits. I'm not actually using `listRemove`; but I am using the others, and I am finding `listReplace` to be particularly useful.

If you would like to take a look at my in-progress code, the brick-related stuff is mostly in the `Cli` modules:
https://github.com/hallettj/locutoria/tree/master/Network/Mail/Locutoria/Cli

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -28,11 +28,13 @@ library
     Brick.Util
   other-modules:
     Brick.Prim.Internal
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Thanks!

PR #2, opened 2015-06-22
Splits out `stepVty`, runs single step of event loop

hallettj:
Because I need to stop and start Vty, I unfortunately have to reach pretty deep into your event loop code. This change adds a new public function that runs one step of the event loop. It requires making `RenderState` public (with a `Default` instance, but without constructors). Adding `stepVty` is the best idea that I have come up with to avoid making the internal details of `RenderState` public.

You have probably already found it, but here is a highlight of the spot where I use `stepVty`:
https://github.com/hallettj/locutoria/blob/32e38f48d0368d0811a018d1340c07337332c887/Network/Mail/Locutoria/Cli/Ui.hs#L49-L54

### DIFF
--- a/src/Brick/Main.hs
+++ b/src/Brick/Main.hs
@@ -14,6 +14,7 @@ module Brick.Main
   , supplyVtyEvents
   , withVty
   , runVty
+  , stepVty
 
   , neverShowCursor
   , showFirstCursor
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Check out the latest changes to brick. In particular, the type of the event handler has changed so that you return one of
```
continue s
halt s
suspendAndResume (IO s)
```
where `s` is your application state type.  When you return `continue`, the vty event loop continues. When you return `halt`, the vty event loop halts and returns the specified state to `main`.  When you return `suspendAndResume`, the vty event loop is shut down, the rendering state is saved, and your IO action is run.  When it is done and returns a new application state, that state is loaded as the new application state and the vty event loop resumes as if no interruption occurred.

You can combine that with the use of custom events by running `customMain`; for an example, see `programs/Events.hs`.  I believe the changes I've made up to now cover the functionality you needed when you submitted this pull request.

PR #16, opened 2015-09-02
Make vScrollPage actually scroll vertically

rootzlevel:
A typo in the definition of vScrollPage made it equivalent to
hScrollPage.


jtdaugherty:
Wonderful. Thank you!

PR #18, opened 2015-09-19
Fix selection handling in listReplace

rootzlevel:
The previous implementation of the merge function was buggy and could
for example return out of bounds indices even for non-empty lists
and generally didn't behave as intended.

This new implementation should be better. It now returns -1 if the
replacement list is empty, but this case is excluded on the caller side
in listReplace.

Please review thoroughly, since the code is a little hairy and I'm not sure
 I thought of all the edge cases.

Anyway, it's hard to come up with a correctness specification for this
selection preservation in listReplace, because it depends crucially on
the ability of the diff algorithm to capture the changes correctly. Maybe
 it would be easier to just assume that elements are distinct and select
 the first item, that compares equal to the previously selected one.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -90,6 +90,21 @@ library
                        template-haskell,
                        deepseq >= 1.3 && < 1.5
 
+test-suite brick-tests
+  type:               exitcode-stdio-1.0
+  hs-source-dirs:     test
+  main-is:            TestDriver.hs
+  other-modules:      ListTest
+  default-language:   Haskell2010
+  build-depends:      base,
+                      brick,
+                      tasty,
+                      tasty-hunit,
+                      lens,
+                      vector,
+                      QuickCheck,
+                      tasty-quickcheck
+
 executable brick-visibility-demo
   if !flag(demos)
     Buildable: False
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -237,8 +237,11 @@ listSelectedElement l = do
 -- the corresponding index in `ys`.
 maintainSel :: (Eq e) => [e] -> [e] -> Int -> Int
 maintainSel xs ys sel = let hunks = D.getDiff xs ys
-                        in merge 0 sel hunks
+                        in clamp 0 (length ys - 1) $ merge 0 sel hunks
 
+-- Given (0, sel, diff), computes the value 'sel' would have after 'diff'
+-- was applied. Retuns -1 if the selected item doesn't appear anywhere in
+-- the replacement list.
 merge :: (Eq e) => Int -> Int -> [D.Diff e] -> Int
 merge _   sel []                 = sel
 merge idx sel (h:hs) | idx > sel = sel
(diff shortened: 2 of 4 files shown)


jtdaugherty:
Thank you for the patch!  To be honest, the functionality involved here was contributed by someone else and it's not something I have used yet.  Although there isn't a `brick` test suite yet, this strikes me as something that would benefit greatly from a few tests to capture what we think is the right behavior.  Would you be willing to do that?  If not, I can get to it some time soon.

If this is something you would like to do, then please use the `tasty` library and consider writing some `HUnit`-style tests (some QuickCheck properties might be appropriate, too, but none come to mind at the moment).

rootzlevel:
I'll see what I can come up with :-)

rootzlevel:
Don't worry, I'll rebase the branch at the end. Just want to document the process.

Two tests are failing, the fix for one is easy and will be pushed shortly, the second is more interesting. It shows that `listReplace` can end up selecting an element that compares unequal to the previously selected one, even if an equal one exists in the replacement list. That's what I meant by "depending on the diff algorithm"

jtdaugherty:
Fantastic!

jtdaugherty:
Have you made any progress on getting these tests to pass?

rootzlevel:
Sorry for the confusion, but this particular failing test was never meant to succeed, but should demonstrate a problem with the whole approach to `listReplace`. I'd remove it in the end, but wanted to highlight the problem.

Other approaches have different problems. For example, I use this function in some of my programs:
```haskell
listSimpleReplace :: Eq e => V.Vector e -> List e -> List e
listSimpleReplace elems oldList =
  let selected = flip V.elemIndex elems . snd =<< listSelectedElement oldList
      newSelected = if V.null elems
                       then Nothing
                       else Just $ fromMaybe 0 selected
  in oldList & listElementsL .~ elems & listSelectedL .~ newSelected
```
It's much simpler and more predictable, but if the list contains multiple equal (under Eq) elements, it always selects the first one after the replace operation. Note that the currently failing test would pass with this implementation, but the "multiple equal"-tests would fail.

It's up to you to decide which semantics you want. I'm currently using the function above, since I don't have multiple equal elements in my list.

jtdaugherty:
Looking at this again, I'm actually tempted to remove `listReplace` from the library because of these issues.  Given that there are a number of reasonable interpretations of "preserve the selected index," I'm concerned there isn't an obvious one.  I'd rather pass the decision onto the application developer since that's really an application-level semantics choice that I don't think the library can make.  If a clear winner shakes out and someone wants to contribute it, then I'd consider merging it or suggesting the creation of a `brick-list-extras` package or something.

With that said, I really appreciate the lengths you went to to test this, so thank you!

jtdaugherty:
FYI: https://github.com/jtdaugherty/brick/commit/7e4b6ff0725ac2c21bdeaef8edcbdeea736c869a

rootzlevel:
Yes, 7e4b6ff is the better solution. Thanks

PR #20, opened 2015-10-22
Changed ListDemo list type from List Int to List Char, small bugfix.

Kaligule:
This might improve the understanding of what is a list item and what
is a position. Also in when running the demo it was possible to create
identical items by removing an element form the middle and adding a
new one (wich got the number (length of list), not (smallest unused
element in list)).


jtdaugherty:
Thanks!

Kaligule:
:) Thanks for your great package with the even greater documentation.

PR #26, opened 2015-11-28
renderList documentation

Profpatsch:
It was not clear what the `Bool` does without looking at the source.


jtdaugherty:
Thank you!

PR #29, opened 2015-12-08
Apply hlint hints

markus1189:
I ran `hlint` and applied some of the suggestions, I restricted myself to the ones that are hopefully undisputed ;)  If there's anything that you don't like I'll undo it.

- redundant brackets
- `fromMaybe x mx` instead of `maybe x id mx`
- redundant do
- functor law: `f <$> g <$> m` == `f . g <$> m`
- combine import statements, e.g.
```
import Brick.Types (rowL, columnL)
import Bricket.Types (Widget)
```

changed to

```
import Brick.Types (rowL, columnL, Widget)
```

This reduces the number of hints from `103` to `25`, of which many are of the following type:

```
./src/Brick/Widgets/Border.hs:113:20: Warning: Move brackets to avoid $
Found:
  (Widget Fixed Fixed $ return middleResult) <+> vBorder
Why not:
  Widget Fixed Fixed (return middleResult) <+> vBorder
```

Given that this change is a little more subjective, I chose to ignore these for now.  If you don't mind either way, I would also be happy to apply the remaining hints.

-- 24pullrequests
[](http://24pullrequests.com/)

### DIFF
--- a/programs/LayerDemo.hs
+++ b/programs/LayerDemo.hs
@@ -8,12 +8,9 @@ import Data.Default
 import qualified Graphics.Vty as V
 
 import qualified Brick.Types as T
-import Brick.Types (rowL, columnL)
+import Brick.Types (rowL, columnL, Widget)
 import qualified Brick.Main as M
 import qualified Brick.Widgets.Border as B
-import Brick.Types
-  ( Widget
-  )
 import Brick.Widgets.Core
   ( translateBy
   , str
--- a/programs/SuspendAndResumeDemo.hs
+++ b/programs/SuspendAndResumeDemo.hs
@@ -61,5 +61,5 @@ theApp =
         }
 
 main :: IO ()
-main = do
+main =
     void $ defaultMain theApp initialState
--- a/programs/ViewportScrollDemo.hs
+++ b/programs/ViewportScrollDemo.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Main where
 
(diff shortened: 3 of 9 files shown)


jtdaugherty:
Thanks for the pull request!

There are *lots* of changes in this patch and there are too many for me to cherry-pick the ones I like.  Of the ones made so far, I find many of them objectionable on stylistic grounds, because they introduce pointless churn for no real benefit, or because any code I write in the future is unlikely to be consistent with `hlint`'s recommendations. :)

I would be happy with changes such as import consolidation since I consider that a real oversight, but some of the changes introducing parens or using `unless` or bothering with `f . g <$> foo` are things I would never write myself and actively dislike, and if I came across some of them in my own code I'd probably change them back to whatever I find natural.  So I can't accept this pull request as-is; I'd be okay with a few of the classes of changes made, but not with others.

jtdaugherty:
(And since the changes are all part of one commit, I'm not sure it would be valuable for me to go through all the hunks and comment on which ones are okay with me, although perhaps if we could boil things down to a set of hlint rules that I am always okay with then we could just run those.)

jtdaugherty:
Looking at the list you mentioned in your original message, the ones I would always be happy with are

* fromMaybe x mx instead of maybe x id mx
* redundant do
* combine import statements

markus1189:
Okay I reverted most of the changes, e.g. redundant brackets and the functor law.

I also added a more detailed list of changes per file:

```
- programs/LayerDemo: combine imports
- programs/SuspendAndResumeDemo: redundant do
- programs/ViewportScrollDemo: unused pragma
- src/Brick/Main: redundant do
- src/Brick/Widgets/Center: use fromMaybe
- src/Brick/Widgets/Core: combine imports, redundant do, use >=>, `unless` instead of `when . not`
- src/Brick/Widgets/Internals: redundant '$'
- src/Brick/Widgets/List: use fromMaybe
- src/Brick/Widgets/ProgressBar: use fromMaybe
```

jtdaugherty:
Thank you!  Can you remove the `unless` changes?  Then I'll be happy to merge.

markus1189:
:+1:

jtdaugherty:
Thank you!

PR #30, opened 2015-12-21
Added MonadIO support to EventM.

AndrewRademacher:
When converting EventM from `type` to `newtype` the MonadIO instance was forgotten.


jtdaugherty:
Thanks!

PR #34, opened 2015-12-27
Derive `List` instances: Functor, Foldable, Traversable

markus1189:
I think adding those instances makes it much easier to work with lists, e.g. operations like `length` for `List a` are automatically available via the `Foldable` instances.

Btw, as you might notice by now, I am currently in the process of migrating one of my personal projects from` vty-ui` to `brick`.  I just wanted to let you know that up until now I really like the new design!  Especially that most operations (on widgets etc) are pure ;)

Best,
Markus

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -1,6 +1,9 @@
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TemplateHaskell #-}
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for the changes! Travis-ci reports build failures for GHC 7.8. It looks like it's just a matter of adding some missing imports. Can you take care of that?

Also, I'm glad to hear that your migration from vty-ui to brick is going well.  Please keep me posted on how that works out! :)

markus1189:
Sure thing :+1:

jtdaugherty:
Thanks!

PR #36, opened 2016-01-03
List: add `listClear` and `listReverse`

markus1189:
In theory, it might be concise enough to use the `listElementsL` lens to do this, but I think these two functions nevertheless provide common enough use cases to justify being a part of the API of `List`.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -30,6 +30,8 @@ module Brick.Widgets.List
   , listRemove
   , listReplace
   , listSelectedElement
+  , listClear
+  , listReverse
 
   -- * Attributes
   , listAttr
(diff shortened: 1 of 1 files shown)


jtdaugherty:
(My comments provide some evidence for `listElementsL` modifications not being quite enough for a lot of interesting list operations since some book-keeping needs to be done. :/)

markus1189:
Right, good catch.  Maybe exposing `listElementsL` is not such a good idea given that it is tempting to violate invariants like keeping the selection valid...

jtdaugherty:
Yeah, exposing that does increase the risk of mismanagement. I chose to expose it to give the user more power. I could update the docs to make it clear that messing with the elements may also entail updating other fields of the `List`. I'm concerned that if I don't expose that, I'd have to make the `List` API pretty big to compensate.

jtdaugherty:
Merged. Thanks!

PR #37, opened 2016-01-07
Add 'lookupViewportR'.

ts468:
Function to access a `Viewport` within the rendering function of a `Widget`.

### DIFF
--- a/src/Brick/Types.hs
+++ b/src/Brick/Types.hs
@@ -44,6 +44,7 @@ module Brick.Types
   -- ** Rendering results
   , Result(..)
   , lookupAttrName
+  , lookupViewportR
 
   -- ** Rendering result lenses
   , imageL
@@ -72,6 +73,7 @@ import Control.Lens (_1, _2, to, (^.), (&), (.~), Lens')
 import Data.Monoid (Monoid(..))
 import Control.Monad.Trans.State.Lazy
 import Control.Monad.Trans.Reader
+import Control.Monad.Trans.Class (lift)
 import Graphics.Vty (Event, Image, emptyImage, Attr)
 import Data.Default (Default(..))
 import Data.Functor.Contravariant
@@ -186,3 +188,8 @@ lookupAttrName :: AttrName -> RenderM Attr
 lookupAttrName n = do
     c <- getContext
     return $ attrMapLookup n (c^.ctxAttrMapL)
+
+-- | Given a name, obtain the viewport for that name by consulting
+-- the viewport map in the rendering monad.
+lookupViewportR :: Name -> RenderM (Maybe Viewport)
+lookupViewportR name = lift $ gets (M.lookup name . (^.viewportMapL))


jtdaugherty:
Can you elaborate on what you're trying to do that needed this patch?

ts468:
I want to build a `Widget` that contains a `List` or `Editor` or in general a `Viewport`, and access the state of the `Viewport` within the rendering function of the `Widget`.

For example:

renderFoo foo specialList = 
  Widget T.Fixed T.Fixed $ do
    _ <- render specialList
    Just vp <- lookupViewportR ("specialList" :: Name)
    render $ hBox [ searchListWidget, translate vp foo ]
  where
    translate = \vp -> translateBy $ Location (0, vp ^. vpTop)


Is there a better way to do it? Alternatively, if `viewportMapL` would be exported, then I could use that instead.

jtdaugherty:
This is going to be tricky to do this way, because you'll be accessing the state of the viewport *before* it is ready for you.  The thing is, viewport data is only available *after* rendering; viewport size information is a *by-product* of rendering so you cannot access it at rendering time.  If you do manage to do so, it may very well be stale because it will change once rendering is complete.  This is because viewport size information is a function of the layout algorithms that run during rendering, and only once those completed will the viewport size data be accurate.

jtdaugherty:
It might be helpful to discuss what you want to do in more general terms to see if there is another design approach that will work.  You only have access to the correct viewport state in your event handler, so you could access it there and update your application state to influence further rendering operations.

jtdaugherty:
Since exposing the viewport state violates an important invariant of the library, I'm going to close this -- but feel free to open an issue to work on solving your problem some other way! :)

ts468:
Thank you very much for reviewing and commenting on the change!

But it's not completely clear to me, yet. Isn't it that he function `lookupViewportR` does not change the state in `RenderM`, it only accesses what has already been rendered? So instead of waiting until the next run of `EventM`, which has access to the completely evaluated `RenderM`, I wanted to access intermediate results in `RenderM` that I know won't change anymore. To ensure to only read the state of `RenderM`, the lens `viewportMapL` would not be exported, but just a kind of `getter`. So more or less it's just a way of wrapping a `Viewport` with another `Widget` for which the layout is determined at rendering time. It's the same way as `viewport` works internally, and I don't quite understand how one could mess up any important invariant of the library with it. Especially if the `Widget` that is associated to the `Viewport` that is looked up will not be re-rendered again afterwards in the rendering pipeline. Does that all somehow make sense? :) At least it works quite well!

jtdaugherty:
You're right that `lookupViewportR` doesn't change the rendering state.  That's okay, but that isn't what I think makes this tricky.  To me the trouble is that you have to know when it is *not yet safe* to call `lookupViewportR`.  In your example above, if you called it *before* rendering the `specialList`, sometimes you would get a `Just` result (i.e. if the list had already been rendered in the past) and sometimes you'd get `Nothing` (if you called this on first rendering).  And if you got a `Just` then you'd get stale data.  The way you wrote it is the only valid way to access the viewport state, and in my opinion that makes this too easy to get wrong, i.e., too easy to write code that compiles and seems right and maybe even behaves right at runtime... most of the time.  I'm concerned that this kind of code will result in me getting spurious bug reports and having to put a big warning on the docstring for the function so people are careful with it, and when I have to do that it's usually a sign that the library design could be improved instead. :)

jtdaugherty:
With that said, I'm open to making that function public only once I'm convinced there are no other good design avenues that don't accomplish the same thing.

jtdaugherty:
Do you have an application you're developing that I could build and run to see your use case in action?

jtdaugherty:
Or perhaps just make a minimal demo program that illustrates what you'd like to see.

ts468:
Thank you very much for your time and effort! I've written a small demo program that should illustrate what I want to achieve.

```
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE TemplateHaskell #-}
module Main where

import Control.Lens hiding (elements)
import Data.Maybe (fromMaybe)
import qualified Graphics.Vty as V
import qualified Data.Vector as Vector

import qualified Brick.Main as M
import Brick.AttrMap (AttrMap, AttrName, attrMap)
import Brick.Util (bg, fg)
import qualified Brick.Types as T
import Brick.Types
  ( Widget(..)
  )

import Brick.Widgets.Core
  ( withAttr
  , hLimit
  , vLimit
  , hBox
  , str
  , translateBy
  , (<=>)
  , padRight
  )

import qualified Brick.Widgets.List as L


data ExtendedList = ExtendedList
                  { extension :: Widget
                  , list :: L.List String
                  , height :: Int
                  , width :: Int
                  }

T.suffixLenses ''ExtendedList


renderExtendedList :: ExtendedList -> Widget
renderExtendedList el =
  Widget T.Fixed T.Fixed $ do
    _ <- render extendedListW
    Just vp <- T.lookupViewportR ("extendedList" :: T.Name)
    render $ hBox [ extendedListW, translate vp (extension el) ]
  where
    translate vp = translateBy $ T.Location (0, offset - (vp ^. T.vpTop))
    offset = fromMaybe 0 $ el ^. listL . L.listSelectedL
    extendedListW = hLimit (width el) $ vLimit (height el) (L.renderList (el^.listL) drawElement)
      where
        drawElement :: Bool -> String -> Widget
        drawElement True a = withAttr selectionAttr $ padRight T.Max $ str a
        drawElement False a = str a 

    
instance T.HandleEvent ExtendedList where
  handleEvent e sl = case e of
                      V.EvKey V.KUp [] -> T.handleEventLensed sl listL e
                      V.EvKey V.KDown [] -> T.handleEventLensed sl listL e
                      _ -> return sl


  
drawUi :: ExtendedList -> [Widget]
drawUi el = [ renderExtendedList el ]


appEvent :: ExtendedList -> V.Event -> T.EventM (T.Next ExtendedList)
appEvent el (V.EvKey V.KEsc []) = M.halt el
appEvent el ev = M.continue =<< T.handleEvent ev el


extensionAttr :: AttrName
extensionAttr = "extension"

selectionAttr :: AttrName
selectionAttr = "selection" 

theMap :: AttrMap
theMap = attrMap V.defAttr
    [ (extensionAttr, fg V.brightGreen)
    , (selectionAttr, bg V.brightRed)
    ]
 

app :: M.App ExtendedList V.Event
app = M.App { M.appDraw = drawUi
            , M.appStartEvent = return
            , M.appHandleEvent = appEvent
            , M.appAttrMap = const theMap
            , M.appLiftVtyEvent = id
            , M.appChooseCursor = M.showFirstCursor
            }



initialState :: ExtendedList
initialState = ExtendedList ext l 20 15
  where
    ext = withAttr extensionAttr $ str "Dummy extension" <=> str "spanning multiple lines"
    l = L.list ("extendedList" :: T.Name) (Vector.fromList $ map (\x -> "item " ++ show (x :: Integer)) [1..45]) 1

                
main :: IO ()
main = do
  _ <- M.defaultMain app initialState
  return ()
```

jtdaugherty:
I played around with other ways to implement what you want, but I don't think any of them are as good as just using the function you added.  I merged it, but as part of communicating the tricky nature of the function, I'm going to rename it `unsafeLookupViewport`.  Thanks!

jtdaugherty:
This change is now released as part of `brick` version `0.4`.

PR #38, opened 2016-01-14
Bug fix for visibility requests + allow 'visible' multiple times

ts468:
Hi Jonathan,

please see my small change to your library. Let me know what you think! :)

Many thanks,
Thomas


jtdaugherty:
That is a tricky section of the library, but those changes look okay to me. Thanks!

PR #39, opened 2016-01-19
Remove a nested if/else in listRemove and removes redundant brackets

Kintaro:


jtdaugherty:
Thanks for the patches.  What prompted these changes?

Kintaro:
Tried to make a lazy version of the list as I have a case where it's filled with millions of lines. While I had the original List.hs open, the linter fired.

jtdaugherty:
I'm going to merge this one, but in the future please be aware that I prefer not to merge changes that don't change the functionality of the library.  Changes that only mess with style, indentation, punctuation, etc. don't add much to the codebase. I also think some Haskell linter suggestions are unwise. :) Thanks!

PR #49, opened 2016-04-03
Remove unnecessary type class constraint

### LINKED ISSUE #48: The type class constraint of Brick.Widgets.List.listReplace
`listReplace :: Eq e => Vector e -> Maybe Int -> List e -> List e`

Is there any reason that the element of Vector must have Eq instance? (I have no idea)
I removed the `Eq e` constraint and the library still compiles.

sifmelcara:
fix #48 .
Thanks.


jtdaugherty:
Thanks!

PR #51, opened 2016-04-21
Radio:  new widget proposal

deepfire:
This is a preliminary proposal for a horizontal Radio widget -- please bash it mercilessly!

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -70,6 +70,7 @@ library
     Brick.Widgets.Edit
     Brick.Widgets.List
     Brick.Widgets.ProgressBar
+    Brick.Widgets.Radio
     Data.Text.Markup
   other-modules:
     Brick.Types.TH
--- a/src/Brick/Widgets/Radio.hs
+++ b/src/Brick/Widgets/Radio.hs
@@ -1,0 +1,130 @@
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE OverloadedStrings #-}
+-- | This module provides a simple radio widget. You get to pick the
+-- radio title, if any, as well as its buttons.
+module Brick.Widgets.Radio
+  ( Radio
+  , radioTitle
+  , radioName
+  , radioButtons
+  , radioSelectedIndex
+  -- * Construction and rendering
+  , radio
+  , renderRadio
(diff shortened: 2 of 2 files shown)


jtdaugherty:
Thanks for the patch!  I will take a look at this in the next few days.  As for the build failure, just import `Control.Applicative` explicitly.

jtdaugherty:
I've had some time to take a look at this.  I wrote a demo program (similar to `DialogDemo.hs`) to play around with it.  I'm going to hold off on commenting on the code itself since first I'd like to comment on the functionality.  Usually when I think of "radio buttons" -- even in terminal applications -- I think of this presentation that I usually see:
```
[ ] One
[*] Two
[ ] Three
```
where `Two` is the currently selected option.  When the user interacts, `Tab` or `Shift-Tab` might be used to cycle through the available options.  So since this is what I think of when I think about radio button inputs, I was surprised by the rendering used by your module, which for me was
```
Title: One Two Three
```
because that doesn't make me think of radio buttons as they are usually presented (and I think observing UI conventions will be important here).

My other main concern with this module is that while users may sometimes want to put all available radio button options next to each other in a window as you have done, I expect that just as often that will not be the case, and that the radio button options will be distributed throughout a more complex UI, such as
```
[ ] One          [(text input box here)              ]
  (other stuff)
[*] Two          [(text input box here)              ]
  (other stuff)
[ ] Three        [(text input box here)              ]
```
and, as part of that kind of layout, the `Tab`-driven cycling between inputs will need to *interleave* the radio inputs with inputs the programmer has chosen to lay out nearby.  Your module doesn't provide that kind of control.  I'm not sure how to do what I've described cleanly so I right now I can't propose an alternative, but when widgets go into the core of `brick` I want to make sure that they are as generic as possible to be useful in the most cases.  Right now I think the proposed module provides too specific a feature to go in as it currently stands so we'll have to see if we can think of ways to address the concerns above.

As an aside, the patch looks great and I really appreciate how it completely matches the style used in the rest of the library and has full documentation.  So thanks!

jtdaugherty:
I'm going to close this due to inactivity.  If you decide to revive this, feel free to re-open and we can continue discussion. Thanks!

PR #54, opened 2016-05-15
renaming Data.Vector in ListDemo.hs

ahgilbert:
In working with some sample programs, I was confused trying to find Data.Vector.cyan for a few minutes. It turned out that ListDemo.hs does qualified imports of both Data.Vector and Graphics.Vty as V. I thought this might save someone else a few head scratches.

Thanks for the library; it looks like exactly what I need!


jtdaugherty:
Yikes! Thanks for the fix.

PR #55, opened 2016-05-16
Add lower bound for microlens package

markus1189:
Prior to 0.3.0.0 e.g. `Microlens.Internal.Field1` was not yet available and the build fails with

```
    src/Brick/Types/Internal.hs:34:29:
        Module ‘Lens.Micro.Internal’ does not export ‘Field1’
    
    src/Brick/Types/Internal.hs:34:37:
        Module ‘Lens.Micro.Internal’ does not export ‘Field2’
```

To solve this I added a lower bound for `microlens`: `microlens >= 0.3.0.0`.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -81,7 +81,7 @@ library
                        transformers,
                        data-default,
                        containers,
-                       microlens,
+                       microlens >= 0.3.0.0,
                        microlens-th,
                        vector,
                        contravariant,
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thank you!

jtdaugherty:
This fix is now released in `0.6.1`.

PR #57, opened 2016-05-22
Derive Functor instance for 'Next'

markus1189:
Missed that one recently and though adding it would be nice.

Btw, you did an awesome job with the `brick` redesign of `vty-ui` I am enjoying the `event-loop` + `pure state` very much :) :+1:

### DIFF
--- a/src/Brick/Types/Internal.hs
+++ b/src/Brick/Types/Internal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 module Brick.Types.Internal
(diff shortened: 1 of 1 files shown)


jtdaugherty:
I kinda can't believe I missed `Functor` there, and `Applicative` needs to be added, too.  Thanks for your compliment! I'm glad you're getting a lot out of it. :)

jtdaugherty:
(On second thought, `Applicative` might not be a good idea yet. `Halt` doesn't behave like you might expect so it would be possible to do things like `f <$> Halt s <*> Continue s2` and that would be bad.)

jtdaugherty:
This change is now released: http://hackage.haskell.org/package/brick-0.6.4

markus1189:
You have a really short time-to-market ;) Awesome :+1:

PR #71, opened 2016-07-31
Bind Ctrl-U to delete the whole text in edit widget

hpdeifel:
This is a standard readline binding

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -88,7 +88,7 @@ library
                        vector,
                        contravariant,
                        text,
-                       text-zipper >= 0.2.1,
+                       text-zipper >= 0.5,
                        template-haskell,
                        deepseq >= 1.3 && < 1.5
 
--- a/src/Brick/Widgets/Edit.hs
+++ b/src/Brick/Widgets/Edit.hs
@@ -49,6 +49,7 @@ import Brick.AttrMap
 -- * Ctrl-d, Del: delete character at cursor position
 -- * Backspace: delete character prior to cursor position
 -- * Ctrl-k: delete all from cursor to end of line
+-- * Ctrl-u: delete all from cursor to beginning of line
 -- * Arrow keys: move cursor
 -- * Enter: break the current line at the cursor position
 data Editor n =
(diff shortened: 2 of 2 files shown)


seagreen:
I usually think of Ctrl-U as being the hotkey for "undo", but maybe standards are different for TUIs.

jtdaugherty:
@hpdeifel - thanks for the pull request.  Is this binding modeled after a particular environment?  Brick's editor widget is intended to provide keystrokes common to environments like `bash`, `readline`, and `emacs`.  In `bash`, for example, `C-u` does not delete all text in the line (as suggested by the pull request title).  From `man bash`, under `Killing and Yanking`:
```
       unix-line-discard (C-u)
              Kill  backward  from  point  to  the beginning of the line.  The
              killed text is saved on the kill-ring.
```
And in `emacs` the `C-u` binding is a command prefix.  Since this binding does different things in all of these environments, I'm concerned that adding it with the behavior implemented here will be confusing.

hpdeifel:
@jtdaugherty You're right, I only checked the behaviour of Zsh and assumed that it was the same in readline. I can change it to kill from the cursor position to the beginning of the line to match bash/readline. Both are derived form emacs keybindings and they chose to bind C-u differently than in emacs. So I think we don't have to worry too much about C-u being the universal prefix there.

@seagreen Which TUI uses C-u as undo? GUIs use C-z which of course is already bound in the terminal.

jtdaugherty:
@hpdeifel Okay, changing it to kill to beginning of line sounds good to me. I don't want brick's editor to be emacs or anything like that, but I do want to be sure that it has unsurprising behavior. :)

seagreen:
@hpdeifel Wow, you're totally right. I have got to step out of a customized world more often, I'm getting confused about what are the actual defaults.

hpdeifel:
Unfortunately `killToBOL` is missing from `Data.Text.Zipper`. Should I implement it there?

jtdaugherty:
Yes, please.

hpdeifel:
Done: jtdaugherty/text-zipper#5

jtdaugherty:
Thanks! Now that `text-zipper` includes the required functionality in `0.5`, can you also update its version bound to `>= 0.5` in `brick.cabal`?

hpdeifel:
done

jtdaugherty:
Thank you!

Issue #72, opened 2016-08-21
`Eq` instances for Location and Cursor


geekingfrog:
I'm having to compare location quite often, and I'd like to avoid to define an orphan instance.
I can do the PR if you want (I believe ghc can derive `Eq` correctly anyway).
Any reason it's not there ?
(disclaimer, I'm fairly new to haskell)

jtdaugherty:
I don't think there's any reason those aren't there.  If you want to write up a patch, I'd be happy to consider it.  Thanks!

jtdaugherty:
I'm closing this due to the merged pull request.  Thanks for the patch!

PR #73, opened 2016-08-22
Added a bunch of deriving Eq

geekingfrog:
As mentioned in #72 I added a bunch of `deriving Eq` for most of the types exported by the library.

### DIFF
--- a/src/Brick/AttrMap.hs
+++ b/src/Brick/AttrMap.hs
@@ -79,7 +79,7 @@ instance IsString AttrName where
 -- | An attribute map which maps 'AttrName' values to 'Attr' values.
 data AttrMap = AttrMap Attr (M.Map AttrName Attr)
              | ForceAttr Attr
-             deriving Show
+             deriving (Show, Eq)
 
 instance Default AttrMap where
     def = AttrMap def mempty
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Thanks for the patch!  Some of the instances you added make sense for simpler types like `ViewportType`, but I'm not sure about others.  For example, I think that if you're ever in a situation where you need to compare `AttrMap` values for equality, there's something else that could be done instead.  And you should never be comparing `Context` values for equality because there is only one.  In general adding instances implies that certain usage patterns are anticipated, and I wan't to discourage usage patterns that rely on using equality with some data structures.  Were those instances added out of a need for those specific operations?  Or were they added out of a desire for completeness?

Here are the instances I think are worth keeping in the patch:

* `VisibilityRequest` `Eq`
* `ViewportType` `Eq`
* `Direction` `Show`, `Eq`
* `Location` `Eq`
* `CursorLocation` `Eq` (although here, too, it's unlikely that you'll ever need equality on these)
* `BorderStyle` `Eq`

geekingfrog:
You're right about implying usage through the types. I added some more for the sake of consistency than any real need, but you have a good point.

I kept the `Eq` instances for the types you mentionned above and removed the other (namely, `AttrMap`, `Viewport` and `Context`). (ff325d978679cddb18609d968de7885c65f3abb8)

Thanks for the feedback.

jtdaugherty:
Okay. Thank you!

PR #77, opened 2016-09-08
Strawman implementation of parametric `Editor` widget

aisamanra:
This isn't a fully backwards-compatible change, but the only thing that needs to be modified for programs involves adding a type parameter, which will be either `String` or `Text`. There's only one example that needed this change, too, which was of course EditDemo.hs. I don't know if this is the actual desired API, but it's a pretty straightforward set of changes to make, so I figured I'd put in the PR like we talked about.

### DIFF
--- a/src/Brick/Widgets/Edit.hs
+++ b/src/Brick/Widgets/Edit.hs
@@ -12,7 +12,7 @@
 -- events that should suffice for most purposes; see the source for a
 -- complete list.
 module Brick.Widgets.Edit
-  ( Editor(editContents, editorName, editDRawContents)
+  ( Editor(editContents, editorName, editDrawContents)
   -- * Constructing an editor
   , editor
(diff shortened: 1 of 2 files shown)


src/Brick/Widgets/Edit.hs
```diff
@@ -12,9 +12,10 @@
 -- events that should suffice for most purposes; see the source for a
 -- complete list.
 module Brick.Widgets.Edit
-  ( Editor(editContents, editorName, editDrawContents)
+  ( Editor(editContents, editorName, editDRawContents)
```
jtdaugherty: Typo here.

aisamanra:
…well, that was weird. I was sure I built it before putting in the PR. Well, typo fixed.

jtdaugherty:
Thanks!

Issue #80, opened 2016-09-15
Expose `nextButtonBy` for `Dialog` widget


geekingfrog:
This is a feature request to expose another way than `handleDialogEvent` to change the state of a dialog widget. It would be nice to have support for arrows as well as tab/s-tab, and maybe something else.

So maybe directly expose (and document) `nextButtonBy` or expose simplified functions like `nextChoice` and `prevChoice` of type `Dialog -> Dialog` to handle that manually.
What do you think ?

jtdaugherty:
I'd be happy to do both or accept patches similarly. Are there other bindings than arrow keys that you think would be useful?

geekingfrog:
I think it would be better to expose generic functions instead of trying to guess what bindings could be useful. I'll work on a patch this week end.

jtdaugherty:
The `Dialog` module doesn't expose much because it doesn't really do very much anyway; the idea with that module was to provide something that would be useful for simple cases, assuming that more complex needs would be met by just implementing something custom. At any rate, if exposing a bit more of the internals would make it more useful, I'm willing to consider it. Thanks!

jtdaugherty:
Now that your pull request is merged, if there's nothing else to be done on this ticket, please close it (or let me know what else needs to be done). Thanks!

geekingfrog:
That was swift ! Thanks.

PR #81, opened 2016-09-18
Add support for ⬌ in dialog

geekingfrog:
In response to #80, after some thoughts, it's probably better to keep the module simple and just add support fort left and right arrow for the dialogs.


jtdaugherty:
Thanks!

Issue #84, opened 2016-10-10
Providing a frp interface


lspitzner:
Looking at the `App` datatype, it seems to me that the interface could be expressed more concisely using some frp library such as `reactive-banana` or `reflex`. (And for users that otherwise already use some frp approach it needs a nasty layer between frp and the current state-passing interface.)

Has this been considered before? Do there exist any forks or other vty-reverse-deps that go in this direction?

jtdaugherty:
I don't know of any existing forks or applications that take this approach, and I haven't considered it before. Although I'm not open to rewriting brick to do this, I would be happy to work with anyone who wants to write an adapter library to expose an FRP interface.

PR #85, opened 2016-10-12
Add reactive-banana interface [draft]

lspitzner:
I have made an attempt at #84; directly as a fork because it seemed easy enough to create a rather shallow wrapper around brick functionality.

Work in progress, I have not tested everything (suspend/resume). Seems to work well enough already, but I am thankful for any suggestions. The `brickNetwork` entry point needs some good documentation; apart from that the only things I dislike are reactive-banana specifics: I currently use IORefs for certain state and require a reactimate-roundtrip. For both I am unable to see how to improve. Otoh the effect on performance probably is negligible and the user-interface is not affected.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -60,6 +60,7 @@ library
     Brick.AttrMap
     Brick.Focus
     Brick.Main
+    Brick.MainBanana
     Brick.Markup
     Brick.Types
     Brick.Util
@@ -90,7 +91,9 @@ library
                        text,
                        text-zipper >= 0.7.1,
                        template-haskell,
-                       deepseq >= 1.3 && < 1.5
+                       deepseq >= 1.3 && < 1.5,
+                       reactive-banana,
+                       stm
 
 executable brick-cache-demo
   if !flag(demos)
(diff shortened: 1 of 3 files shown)


jtdaugherty:
I can't really comment on the FRP content in the patch, but it looks like it would probably be totally suitable as a separate package. (I'm assuming this was just done as a proof of concept; I'm not going to merge it into `brick` proper.) But I'm glad you were able to get it to work!

lspitzner:
Yes, this was mostly a proof of concept and doing things in-package was simply the quickest way to get this going.

There is one problem though with separation - `Brick.Types.Internal` and `Brick.Widgets.Internal` are not exposed. _If_ I decide to publish a separate package, that would require some resolution.

jtdaugherty:
Is there a way to write this wrapper without using those internals directly? I'd really like to avoid exposing either of those modules.

lspitzner:
The references to `Next` and `EventState` can be eliminated I think. But that still leaves `RenderState` and `renderFinal` where I currently see no way.

jtdaugherty:
In looking more closely at the diff, I'm concerned that there's enough duplication of code from brick's internals that any third-party package that provides this can drift from brick's implementation and thus provide a different behavior than one would expect from using brick itself. (In particular I'm thinking of the `render` function.)

Also, having being totally unfamiliar with FRP in general, I feel hesitant about this integration because I have no idea how brick's design choices are going to interact with FRP idioms. Brick wasn't designed to be used as an FRP library, so I'm concerned that trying to use it that way will be awkward. You could probably assess this if you know brick well enough.

I'm not comfortable exporting any of the Internal module contents, though - but if you explore this more and find that some tweaks to already-exported APIs would be helpful to make this work, I'd be willing to consider those.

PR #91, opened 2016-10-31
Make FocusRing a Functor.

seagreen:


seagreen:
I'm using this to generalize the type of a `FocusRing` before calling `withFocusRing` on it.

Concretely, I have a function `widgetName :: MyConcreteWidgetType -> a` and am using it like this:
```
F.withFocusRing (widgetName <$> myFocusRing) E.renderEditor myEditor
```

Without the `<$>` I end up with:
```
Could not deduce (Brick.Widgets.Core.Named (E.Editor String a) MyConcreteWidgetType
```

jtdaugherty:
It looks like instead of putting name values in your focus ring initially, you're putting something else (here `MyConcreteWidgetType`). Is that right? The intended usage is to put names in the ring, which is why there is no `Functor` instance.

jtdaugherty:
(And to be clear, I'm not suggesting you shouldn't be doing that; I'd just like to understand more about your use case, and whether there's a use I had overlooked.)

seagreen:
> (And to be clear, I'm not suggesting you shouldn't be doing that; I'd just like to understand more about your use case, and whether there's a use I had overlooked.)

No worries, that's why I left the comment -- I wanted to make sure I needed `Functor` for a good reason.

I was originally putting `a` values in the focus ring. Unfortunately this meant passing the `widgetName :: MyConcreteWidgetType -> a` function through a lot of code, because at various parts of the code I want to focus specific widgets, and I couldn't do that without a way to turn that widget name into an `a`.

Additionally when I'm pattern matching to draw the Widget, if I'm working on `MyConcreteType` I don't have to pass an `a -> Maybe MyConcreteType` type function, which is nice (all of this is happening in a library that doesn't know what the eventual type of `a` will be).

This is just a convenience issue though, so I understand if it's not worth adding the Functor instance.

(One thing that would semantically matter would be if there was a `NonEmptyFocusRing` based around `NonEmpty` instead of `[]`. Right now I have some partial functions in my code where I'm not sure what to do if `F.focusGetCurrent` returns `Nothing`. This might be a topic for a separate issue though)

jtdaugherty:
Okay. I don't mind at all about adding this - I just wanted to make sure I understood the need. Thanks!

PR #92, opened 2016-11-01
List is Traversible

kquick:
Simplification of List implementation because it is an instance of Traversible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,60 @@
 Brick changelog
 ---------------
 
+ * Make FocusRing a Functor.
+
+0.12
+----
+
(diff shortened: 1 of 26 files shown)


jtdaugherty:
It's actually `Foldable` that provides `length`. But this build is failing because `Foldable` is not in the `Prelude` on GHC 7.8.4.

I'm not inclined to accept this change because it is purely internal to the List module, rather than adding anything to its interface. Although use of the `Foldable` instance is okay, I consider that part of the external interface of the type rather than the internal interface that I want to use in the `List` module.

kquick:
Agreed.  It was simply an attempt at simplification, but I was unaware of the lack of Foldable in GHC 7.8.4.  Sorry for the noise.

PR #98, opened 2016-11-30
Enhance MouseDemo to show interaction between clickable and viewport.

kquick:


programs/MouseDemo.hs
```diff
@@ -51,6 +56,18 @@ buttonLayer st =
...
                str (if wasClicked then "<" <> label <> ">" else label)
 
+editLayer :: St -> Widget Name
+editLayer st =
+  B.border $
+  C.hCenterLayer $
+  vLimit 8 $
+  -- n.b. if clickable and viewport are inverted here, click event
+  -- coordinates will only identify the viewable range, not the actual
+  -- editor widget coordinates.
```
jtdaugherty: The editor already embeds a viewport, so this use of `clickable` is still outside that viewport here. I would therefore expect any click events to still be just in terms of the viewed area, not the actual editor coordinates. Are you seeing different behavior than I'm describing? (In particular, I'd expect this to still report click events within the viewed area even if the editor was scrolled off to the right, say.)

jtdaugherty:
Thanks for the patch! I left a comment. The use of an editor here might not demonstrate what you want to show very well because of the issue I mentioned. But a non-editor use of a viewport would definitely show how clicks can be detected inside a viewport.

programs/MouseDemo.hs
kquick: Yes, the editor is probably a confusing example for that reason.  I should (and probably will) change this to a viewport of a vbox of str widgets which is better.

With the current code, if you invert the next two lines, the click events are reported for the viewport widget wrapper (i.e. the row is always 0-7, regardless of the scrolling of the viewport contents).  In the order shown here, if the viewport contents are scrolled down by 3 lines then the click reporting row will be in the range 3-10.

I will check what the functionality is if the EditedView viewport is removed in favor of the Editor's internal viewport.

kquick: As followup, using the editor's internal viewport, the clickable row reporting is always 0-7, regardless of the scrolling from the viewport (which matches the behavior I observed here as well).  That's not necessarily optimal for determining which line of text was clicked on.. that's where the getEventWidgetLocation helper function could be useful (which was supposed to be a separate pull request; git surprised me, and I cancelled this pull request to submit two independent cleaner requests).

src/Brick/Main.hs
```diff
@@ -316,6 +319,48 @@ findClickedExtents pos = EventM $ asks (findClickedExtents_ pos . latestExtents)
...
+getEventWidgetLocation widget screenCol screenRow =
+    do mExtent <- lookupExtent (getName widget)
+       case mExtent of
+         Nothing -> return Nothing
+         Just e@(Extent _ upperLeft _) ->
+             if clickedExtent (screenCol, screenRow) e
+             then let widgetRow = screenRow - upperLeft^.locationRowL
+                      widgetCol = screenCol - upperLeft^.locationColumnL
+                      widgetLoc = Location (widgetCol, widgetRow)
+                  in do mView <- lookupViewport (getName widget)
```
kquick: I agree about the assumption of the naming, but the alternative is to explicitly pass the names which makes the interface more awkward.  My thinking was that if the user doesn't use the naming described in the documentation comment that they can use this function as a guide instead to implement a version more appropriate for their needs.

The use of this function is not limited to the Editor widget: in fact, I wrote it for use in my own Widget but thought it might be more generally applicable.

PR #100, opened 2016-11-30
Added getEventWidgetLocation for widget-relative mouseclick locations.

kquick:


jtdaugherty:
Instead of this change I'd like to patch `renderEditor` instead as described in my comment at
https://github.com/jtdaugherty/brick/pull/98/commits/4a20d4813aea844d834e097d1e45f84ab813982c#diff-02db024756acc3cf510c10bb7195684cR357

jtdaugherty:
Okay, I've pushed a patch to do what I described in https://github.com/jtdaugherty/brick/commit/c72caa24d9c484320e903511eb522d7300b09496. In the specific case of editors that makes this function unnecessary since the click event you get will already be in local coordinates in the editor. That, combined with the fix I have planned for #101, should improve matters in this area!

PR #107, opened 2016-12-31
Use DList to avoid repetitive Data.List.++

maoe:
When there are many (>= 10k) items in a `vBox`, the list concatenation in `renderHis` can be really expensive. I have no proper benchmarks but I see a significant latency improvement in [maoe/viewprof](https://github.com/maoe/viewprof) with a large profiling report.


jtdaugherty:
Nice catch. Thanks for a patch to improve this!

Issue #106, opened 2016-12-30
Consider using bounded FIFO instead of unbounded


jchia:
`Control.Concurrent.Chan.Chan` is unbounded. `Control.Concurrent.STM.TBQueue.TBQueue` is bounded.

Someone thinks that unbounded FIFOs are usually a bad idea and I tend to agree. http://lorepub.com/post/2016-12-17-Haskell-Pitfalls

Currently, my own app avoids unboundedness by using two channels, one for each direction, one channel just for ACK. On the other channel, I don't send another item until I have received an ACK. If I use a bounded channel of capacity 1, I don't need the ACK mechanism and I just need to use one channel.

jtdaugherty:
I would be happy to consider a patch migrating the library to, say, `TBQueue`.

jchia:
`Control.Concurrent.Chan.Chan` and `Brick.Main.defaultMain` et al operate in `IO` whereas `Control.Concurrent.STM.TBQueue` operates in `STM`. Thus, I think switching to `TBQueue` entails both the library and the user using `atomically :: STM a -> IO a` for reading and writing the channel. Seems a bit cumbersome.

On the other hand, `unagi-chan` operates in `IO`, but I don't know how popular or stable `unagi-chan` is.

Thoughts?

jtdaugherty:
Since `stm` is going to be available everywhere, I feel better about adding the `stm` dependency. From an API perspective, I want this hidden behind a brick API so users don't have to care that it's implemented using STM. So, something like this (although I want to think more about naming it):
```
data BrickChan a = BrickChan (TBQueue a)

newBrickChan :: IO (BrickChan a)
writeBrickChan :: BrickChan a -> a -> IO ()
readBrickChan :: BrickChan a -> IO a
```

jtdaugherty:
Also, I forgot to include a bound parameter to `newBrickChan`.

jchia:
It turns out that `customMain` relays user events from one channel to another in a special thread. Thus, whatever capacity the user sets on the input `mUserChan` has diminished effect on flow control. E.g. even if he uses a capacity of 1, if the downstream channel has a capacity of 2, there can be up to 3=1+2 outstanding user events.

Unfortunately, AFAIK, there's no built-in facility supporting blocking read from multiple channels in `TBChan` a la Unix `select`. Otherwise, we could put terminal events and user events in separate channels.

jchia:
I think I found a way to read from two channels, using `orElse`.

[...]

PR #111, opened 2017-01-07
Switched to using TBQueue (bounded) instead of Chan (unbounded)

jchia:
#106

In `readBrickEvent` I'm currently prioritizing user events over vty events. I think this means that vty event processing can get starved by user event processing. Is this OK or do you think I should switch the priority (by flipping the argument order) or try to not always prefer one side?

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -58,6 +58,7 @@ library
   exposed-modules:
     Brick
     Brick.AttrMap
+    Brick.BChan
     Brick.Focus
     Brick.Main
     Brick.Markup
@@ -88,6 +89,7 @@ library
                        microlens-mtl,
                        vector,
                        contravariant,
+                       stm >= 2.4,
                        text,
                        text-zipper >= 0.7.1,
                        template-haskell,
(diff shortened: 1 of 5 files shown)


jtdaugherty:
I think Vty events need to take precedence. I suspect that the user's experience of latency will be more dramatic if the latency is on input events rather than custom events.

jtdaugherty:
Thank you for the patch! It looks good to me overall. Some changes that would make merging this easy are:

* Updates to `guide.rst` since that has material on dealing with custom events
* Updates to the `BChan` Haddock strings to use the word "channel" instead of "queue" (for consistency with the type name)

jtdaugherty:
Also, out of curiosity, did you build `brick` with `-f demos` to test this?

jtdaugherty:
Oh, I missed the hunks in the custom event demo before.

jchia:
Yes, I could build the demos. I tried a few of them including `brick-custom-event-demo` and they all looked normal.

jtdaugherty:
Thanks for the updates! The last thing I'd like to see is some text in the guide dealing with the channel bound parameter: the reason it must be provided and the relevant design considerations. Would you be willing to add some text about that?

jchia:
I just added a paragraph on BChan.

jtdaugherty:
Thank you!

jtdaugherty:
This is now released in `0.16`.

PR #113, opened 2017-01-30
add stack config

ababkin:


jtdaugherty:
Thanks for the patch! However, since I do not support `stack`, I cannot accept this patch. I would be happy to update `gitignore`, though.

jtdaugherty:
FYI: https://github.com/jtdaugherty/brick/commit/8034c0015a840fda02e821fbc84ddd71f3462153

ababkin:
Just out of curiosity, why no love for `stack`?
I think the config file can happily co-exist with the regular cabal file, for those who would like to build your project with `stack`

jtdaugherty:
Since I don't use `stack`, I will not be using the configuration file. And if I don't use it, then it will eventually bit-rot. Since I don't want to keep anything in the repository that will eventually break (because I want to keep the promise that I'm doing my best to test everything), I don't want to include it in the repository.

PR #134, opened 2017-08-22
Add type signature to the demo codes

psibi:
While learning any library, the first thing I usually do is get
familirized with the type signatures of the library. I'm adding
singatures to some of the code where I had to consult haddock.

### DIFF
--- a/programs/MarkupDemo.hs
+++ b/programs/MarkupDemo.hs
@@ -22,9 +22,12 @@ import Data.Text.Markup ((@@))
 ui :: Widget ()
 ui = (m1 <=> m2) <+> (padLeft (Pad 1) m3)
     where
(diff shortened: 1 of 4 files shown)


jtdaugherty:
Thanks for the patch! I'm afraid I can't accept it, though. I totally see how adding type signatures was helpful to you in learning the library, and I certainly support adding them if it helps. However, I'm concerned that if I accept this patch, it will make future patches to add type signatures more likely, and I don't want to start adding these type signatures, mostly because I think once someone has learned the API, they don't provide much value. It will also lead to inconsistent style in the demo programs, which I'd like to avoid.

If there is anything I could do to the documentation or demos to make things easier to understand without adding these type signatures, I am happy to discuss it!

Thanks!

psibi:
No issues! Thanks for the excellent documentation.

jtdaugherty:
You're welcome, I'm glad you're finding it useful! Thanks for understanding.

Issue #139, opened 2017-10-01
Add a list handler w/ vi keybindings?


rhofour:
In my own project I'm adding vi keybindings for list movement. Would there be any interest in upstreaming this?

I was thinking something like:
`handleListEventVi :: Ord n => (Event -> List n e -> EventM n (List n e)) -> Event -> List n e -> EventM n (List n e)
`
So instead of: `handleListEvent event list`, you could use `handleListEventVi handleListEvent event list` to add the extra keybindings. Alternatively, these bindings could just be added to handleListEvent.

jtdaugherty:
Which `vi`-style bindings are added by your handler?

rhofour:
Just movement stuff. Up, Down, Page Up, Page Down, Half Page Up, Half Page Down, and maybe start and end. So nothing particularly interesting.

jtdaugherty:
I'd be okay with a patch to `brick` for this provided it's implemented as a separate function as you proposed above.

rhofour:
If it doesn't prove too hard I'll also add: top-of-screen, middle-of-screen, and bottom-of-screen

I don't think these are currently possible to implement with the exposed functions so I'll also probably add another function like listMoveToRelative where listMoveToRelative 0 moves to the first element displayed. Does that also sound reasonable?

rhofour:
Also, I was thinking it made sense to showcase this in brick-list-demo. Let me know if you disagree.

jtdaugherty:
I think that `listMoveToRelative 0` is actually `listMoveToAbsolute`; there is already `listMoveBy` for relative movement, and `listMoveTo` for absolute movement, so those already exist. You can implement `listMoveToTop` and `listMoveToBottom` in terms of `listMoveTo`.

jtdaugherty:
(I now see what you mean by "first element displayed." I'm not sure how valuable bindings for that would be, but the rest would be okay with me.)

jtdaugherty:
As for the list demo, I'd prefer a separate list demo that shows these bindings in use, just so the list demo that I already have can remain as simple as possible and the extra demo can show how it can be extended. I realize the code will be almost entirely the same, but my goal with the demos is to show only one feature at a time.

rhofour:
[...]

PR #140, opened 2017-10-01
Add a list event handler with vi keybindings and a demo.

rhofour:
This adds the vi-keybindings I mentioned in #139 , except the ones for moving around in the visible selection.

ListDemoVi.hs is identical to ListDemo.hs except for line 55 where the vi bindings are added.


programs/ListDemoVi.hs
```diff
@@ -0,0 +1,91 @@
+{-# LANGUAGE OverloadedStrings #-}
+module Main where
+
+import           Control.Monad        (void)
```
jtdaugherty: Please reformat these import statements without the aligning whitespace.

src/Brick/Widgets/List.hs
```diff
@@ -98,6 +99,30 @@ handleListEvent e theList =
         EvKey KPageUp [] -> listMovePageUp theList
         _ -> return theList
 
+-- | Enable list movement with the vi keys with a fallback if none match.
+-- Movements handled: Up, Down, Page Up, Page Down, Half Page Up, Half Page
```
jtdaugherty: Please document which Vi keys correspond with which behaviors in this Haddock.

rhofour: I was on the fence about this since these are all standard bindings, but I can document them.

jtdaugherty: Thanks! Even if they are standard bindings, documenting them makes this handler consistent with the other list documentation and makes this library easier to use for someone who is *not* already a `vi` user.

jtdaugherty:
Thanks!

PR #141, opened 2017-10-06
Add herms to featured projects

JackKiefer:
https://github.com/jackkiefer/herms

We're a growing project that uses a Brick UI as one of our core functionalities. In fact, a screenshot of this UI is prominently featured in our README. :)


jtdaugherty:
Very cool! Thanks!

Issue #144, opened 2017-10-21
Suggestion: Make brick handlers composable


NorfairKing:
First of all: Love your work. Keep it up!

The current way to write brick event handlers is rather limiting when it comes to composing handlers.

What would you think about the following change:

```
data App s e n = {
   [...]
   appHandleEvent :: BrickEvent n e -> Handle s n ()                                    
}
type Handle s n = StateT s (MkNext (EventM n))          

data MkNext f a = MkHalt -- Stop here
                | MkPure a -- Pure value
                | MkCont (f (MkNext f a)) -- ^ Do something, then continue
                | MkSusp (IO (f (MkNext f a))) -- ^ Suspend, do some IO, then continue

```
... and here are the instances I imagine:
```

instance Functor f => Functor (MkNext f) where
    fmap _ MkHalt = MkHalt
    fmap f (MkPure a) = MkPure $ f a
    fmap f (MkCont a) = MkCont $ fmap f <$> a
    fmap f (MkSusp a) =
        MkSusp $ do
            mkn <- a
            pure $ fmap f <$> mkn

instance Applicative f => Applicative (MkNext f) where
    pure = MkPure
    MkHalt <*> _ = MkHalt
    MkPure fa <*> mka = fa <$> mka
    MkCont ffa <*> mka = MkCont $ fmap (<*> mka) ffa
    MkSusp ionfa <*> mka =
        MkSusp $ do
            mna <- ionfa
            pure $ fmap (<*> mka) mna

instance Applicative f => Monad (MkNext f) where
    MkHalt >>= _ = MkHalt
    MkPure a >>= fmkb = fmkb a
    MkCont fa >>= fmkb = MkCont $ fmap (>>= fmkb) fa
    MkSusp iomka >>= fmkb =
        MkSusp $ do
            mka <- iomka
            pure $ fmap (>>= fmkb) mka
```

I'm trying to implement it on my fork, but I'm having some trouble working through `Main.hs`, so I wanted to hear your thoughts first.

jtdaugherty:
Thanks! I'm interested in this, and it comes up with some regularity (and others have requested something like it). I'd like to suggest a similar alternative that I think will be simpler: rather than adding a new monad `Handle` to wrap `EventM`, make `EventM` itself capable of this behavior.

NorfairKing:
I just realised that the StateT on the outside does not make sense (while trying to implement this idea).
Would you like to implement this?
Is there anything I can do to help?

[...]

PR #145, opened 2017-10-21
Making Event handlers composable

NorfairKing:
I tried making the change that I described in #144.

I'm in the process of testing the changes out, so expect some changes to this PR as well.

Please have a really close look at the logic in `Main.hs` because I honestly don't know much about brick/ncurses/vty.

If you like the changes, I'll fix the demos as well, and make the CI pass.


jtdaugherty:
Thanks! I'm still evaluating this, but in the mean time: can you please remove the `stack.yaml` that got checked in?

jtdaugherty:
Also: the original suggestion here was to increase composability of the `EventM` interface. You proposed changes but I never asked: what kind of composability improvement are you proposing? Is it mostly about flow control? That is, are you wanting the following?
```
halt s >> x = halt
```

jtdaugherty:
(Put another way, what is the program you want to write that you can't write with the current interface?)

NorfairKing:
> Can you please remove the stack.yaml that got checked in?

Sure! Make sure to squash when you merge this PR.

> That is, are you wanting the following?

This is a consequence of the change, yes.

> What is the program you want to write that you can't write with the current interface?)

I would like to compose two 'things' that could be used as `appHandleEvent`, ideally as a monadic composition.
Before this change, this was very verbose in the best case, and impossible in the worst case: You cannot compose two `suspendAndResume` handlers.

jtdaugherty:
> I would like to compose two 'things' that could be used as appHandleEvent, ideally as a monadic composition. Before this change, this was very verbose in the best case, and impossible in the worst case: You cannot compose two suspendAndResume handlers, for example.

It would still be helpful for me if you could be more specific. Can you provide an example of an event handler scenario that you are thinking of?

NorfairKing:
> It would still be helpful for me if you could be more specific.

I don't understand how this is not specific.

To give you some more information on my specific use-case.
I want to predefine a whole bunch of actions in a library, and let users of the library compose the event handlers from these small building blocks.

In the case of an editor, you may have `moveToTheRight` and `deleteCharacter` handlers, and then let the users compose the two and bind it to the handler for the 'c' keypress. Then I don't have to write a `moveToTheRightAndDeleteCharacter` function in the library, for each composition that a user may want.

Given that we already have great syntax for monads, and that they are the essence of composition, it makes sense to use a monad for this.

NorfairKing:
In short: You can not compose two functions of type `s -> BrickEvent n e -> EventM n (Next s)` because of the `SuspendAndResume` constructor.

jtdaugherty:
> I don't understand how this is not specific.

By "specific" I meant that specific code examples illustrating the composition difficulty you are trying to address would be helpful. I ask because there is more than one consequence of the changes you made, and since there are multiple ways the current API could be improved, it is not obvious which one you intend. I need to know which one you intend so 1) I can evaluate your changes to see whether they even provide the intended benefit, 2) I can consider alternative approaches that still address your concerns, 3) I can characterize the changes properly in the change log, 4) I can evaluate how your changes dovetail with other changes I have been thinking of making, and 5) I can decide whether the API burden introduced by the changes is outweighed by the benefit.

> In the case of an editor, you may have moveToTheRight and deleteCharacter handlers, and then let the users compose the two and bind it to the handler for the 'c' keypress. Then I don't have to write a moveToTheRightAndDeleteCharacter function in the library, for each composition that a user may want.

Thanks for that example. What are the types of your functions? For example, I would expect either
```
moveToTheRight :: Editor t n -> Editor t n
deleteCharacter :: Editor t n -> Editor t n
```
or
```
moveToTheRight :: Editor t n -> EventM (Editor t n)
deleteCharacter :: Editor t n -> EventM (Editor t n)
```
but both of those are composable already.

NorfairKing:
> What are the types of your functions? 

Neither of your suggestions. In order for them to be usable as handlers _by_themselves_, they have to be of type `s -> BrickEvent n e -> EventM n (Next s)`, which is not composable.

jtdaugherty:
> In order for them to be usable as handlers by_themselves, they have to be of type s -> BrickEvent n e -> EventM n (Next s), or s -> EventM n (Next s), which is not composable.

That's true. Is it the case that you want to write editor keybinding handlers that have the power to kill the event loop (by calling `halt`) and are still composable?

NorfairKing:
> That's true. Is it the case that you want to write editor keybinding handlers that have the power to kill the event loop (by calling halt) and are still composable?

Yes, but not just halt, also suspend.

NorfairKing:
That is not the main issue though.
Event handlers are not composable BECAUSE of these halt and suspend.
I'm not saying I want to compose halt and suspend, but these are the reasons that it's not possible.

jtdaugherty:
> Event handlers are not composable BECAUSE of these halt and suspend.

Yeah, that makes sense. But I want to distinguish between two things:
 * Functions that return `EventM ... (Next ...)`. Let's call these "event handlers" because they not only take events but influence program control flow.
 * Other functions which run in `EventM`. These can observe/modify some of the brick environment but they're otherwise pure and can't influence program control flow.

Event handlers aren't composable because they return a `Next`, and this is deliberate because I don't want non-top-level handlers calling `halt` (just as in C it isn't good practice for library functions to call `exit()`). In the case of the editor transformations, that's why I recommend avoiding `Next` entirely (especially if you don't need it, and from what you're saying, it sounds like you don't).

I do see your point about not being able to compose suspend/resume handlers, and that is something I can probably just pull out of `Next` entirely - I'll give it some thought.

NorfairKing:
> I want to distinguish between two things:

I would like to combine these two, for ease of use.

Otherwise you still have to manually string along the state as well, which is the main source of accidental complexity right now.

> (just as in C it isn't good practice for library functions to call exit()).

Now you allow for `liftIO exitFailure`, but not `halt` ...

jtdaugherty:
> I would like to combine these two, for ease of use.
> 
> Otherwise you still have to manually string along the state as well, which is the main source of accidental complexity right now.

Well, if we're just talking about how the state is passed along, that's a separate issue that can be solved separately (by making `EventM` a state monad over `s`) and has its own design trade-offs. But at present I'm still not open to changing how control flow works (specifically, `halt`). But as I said before, I can look into making `suspendAndResume` easier to use without `Next`.

> Now you allow for liftIO exitFailure, but not halt ...

Yeah, I can't disallow anyone from doing that. But the use of `MonadIO` in `EventM` is deliberate, too: event handlers absolutely need to be able to do I/O (network, disk, etc), otherwise they're not very useful. And the difference between users running `liftIO exitFailure` and `halt` is that the former goes around my library and the latter is provided by it, with an implicit message that it is part of the usage model.

NorfairKing:
> I'm still not open to changing how control flow works.

I'm not suggesting to change control flow. I'd just like the DSL to change for ease of use.

NorfairKing:
In other words: I don't really see the problem. Handlers should be able to halt the control flow, why only let them do that at the end if it comes at the cost of composability?

jtdaugherty:
> Handlers should be able to halt the control flow, why only let them do that at the end if it comes at the cost of composability?

From my earlier comment:

> ... this is deliberate because I don't want non-top-level handlers calling halt (just as in C it isn't good practice for library functions to call exit()). In the case of the editor transformations, that's why I recommend avoiding Next entirely (especially if you don't need it, and from what you're saying, it sounds like you don't).

The current patch changes control flow behavior (by folding halt etc. into `EventM`). (I know originally requested that in our other ticket discussion, but it's becoming clearer to me now why I hadn't done it in the past.) I appreciate the desire to make the API easier to use, and the things I'm willing to do are:

* consider making `EventM` a state monad over the application state to avoid state-passing (or consider adding a new monad that does this, to preserve backwards compatibility), and
* investigate whether `suspendAndResume` can be removed from `Next` to make it easier to compose.

I'm aware that those ideas don't change the ability to compose handlers in the way you requested with the editor case above, but I've explained why that's deliberate. I've also suggested alternative APIs to the handlers you mentioned that are more composable.

NorfairKing:
> I'm aware that those ideas don't change the ability to compose handlers in the way you requested with the editor case above, but I've explained why that's deliberate.

Would it be possible to expose two API's, one where the handlers are composable (and more powerful wrt halt), and one where they aren't (the way you describe), just like you have `simpleMain` and `defaultMain`. Maybe call it `extraCrazyMainForWeirdos` or something.

I really don't want to have to maintain a fork, just to be able to use the library. I'd prefect to have an 'unsafe' part of the library that I can use (because I need it), while most use-cases don't have to use that part, instead of limiting all uses.

jtdaugherty:
I don't think it's going to be possible to maintain two different interfaces, and I'd prefer not to. As for you maintaining a fork, I'd certainly prefer to avoid that, too! However, while I understand there are some things that can be improved here, I'm surprised you consider this a show-stopping issue for the library based on my experience writing non-trivial programs with the library and based on my experience hearing from many users for whom this was not a show-stopping issue. It's possible they didn't tell me, but I doubt it.

> I really don't want to have to maintain a fork, just to be able to use the library.

I recognize that this might be significant enough to make you rule out using the library, but in situations like this I often find it very helpful to get a look at the application in question. This helps me see the problem(s) in action and it may also be an opportunity for me to help you improve your design without changing the library and without abandoning it entirely, or it may help me to see a spot where your use of the API is making things harder rather than easier.

Our interaction here has been difficult for me. In general I've gotten the impression that you might consider this to be an obvious flaw with an obvious solution, and I've felt frustrated when trying to get answers to questions that would help me work with you better and understand your situation. Your comments have given me the impression that you might find it annoying to have to provide detail or rationale about what you're proposing, and that doesn't make me feel motivated to keep trying to get to the bottom of the issue. That makes me feel frustrated enough to give up, and I'd rather not do that.

In the future, if you keep working with Brick and have more changes to propose, it would be more helpful for me if we can make sure that proposals begin with a detailed explanation of the concern to be addressed, rather than a solution (as in this case) and that an application program or code sample that suffers from the problem can be made available. If you have an application that you can share in this case, I'd be happy to take a look!

NorfairKing:
> I don't think it's going to be possible to maintain two different interfaces, and I'd prefer not to. As for you maintaining a fork, I'd certainly prefer to avoid that, too! However, while I understand there are some things that can be improved here, I'm surprised you consider this a show-stopping issue for the library based on my experience writing non-trivial programs with the library and based on my experience hearing from many users for whom this was not a show-stopping issue. It's possible they didn't tell me, but I doubt it.

Let me break down the different facets of the composability and the current problems.

- `Halt`
- `suspendAndResume`
- Passing around application state manually

For composability, passing around application state is not a problem, it would just be 'nicer' if `EventM` was a `MonadState` for the application state.

`suspendAndResume` is a problem for composability, because now you can never compose a `suspendAndResume` with anything after that, because we want to be able to rely on the result in what follows.

`halt` is a problem for composability because you (and I agree with you) would prefer not to have `halt >> x = halt`, or at least not make that specific composition possible. 

Here is a suggestion:

- Build in `suspendAndResume` into `EventM`, so that the handlers become:

```
data PossiblyHalt = Halt | Don'tHalt
data App s e n = {
    appHandleEvent :: s -> BrickEvent n e -> EventM n (s, PossiblyHalt)
}
```
This change cannot be bolted-on if we don't make it. 

- (Optionally) make `EventM n s` a `MonadState`. This can be bolted-on if necessary, if you don't want this change. Then we would have:

```
data PossiblyHalt = Halt | Don'tHalt
data App s e n = {
    appHandleEvent :: BrickEvent n e -> EventM n s PossiblyHalt
}
```

- Don't build in `halt` into the `EventM`. This maintains your idea of 'don't call halt in arbitrary places'.
If anyone really needs that still, they can bolt this on by wrapping `EventM n s (a, PossiblyHalt)` into a newtype that can then have the appropriate `Monad` instance.

> >    I really don't want to have to maintain a fork, just to be able to use the library.
>
> I recognize that this might be significant enough to make you rule out using the library, but in situations like this I often find it very helpful to get a look at the application in question. This helps me see the problem(s) in action and it may also be an opportunity for me to help you improve your design without changing the library and without abandoning it entirely, or it may help me to see a spot where your use of the API is making things harder rather than easier.

I could add you to my project, but it's not open-source and will not always be possible in general.
Please let me know if I should add you.

> Our interaction here has been difficult for me. In general I've gotten the impression that you might consider this to be an obvious flaw with an obvious solution, and I've felt frustrated when trying to get answers to questions that would help me work with you better and understand your situation. Your comments have given me the impression that you might find it annoying to have to provide detail or rationale about what you're proposing, and that doesn't make me feel motivated to keep trying to get to the bottom of the issue. That makes me feel frustrated enough to give up, and I'd rather not do that.

This is really fair. Thank you for bringing it up.
I have spent more than a day of my weekend on this change, so everything about it seemed 'obvious' at the point where you asked those questions.

> In the future, if you keep working with Brick and have more changes to propose, it would be more helpful for me if we can make sure that proposals begin with a detailed explanation of the concern to be addressed, rather than a solution (as in this case) and that an application program or code sample that suffers from the problem can be made available. If you have an application that you can share in this case, I'd be happy to take a look!

Also fair enough.
I hope that my above suggestion is more aligned with what you had in mind.

jtdaugherty:
> For composability, passing around application state is not a problem, it would just be 'nicer' if EventM was a MonadState for the application state.

Yeah, and I've seen other applications do this exact thing when they need it. I prefer that approach (i.e. not doing it in Brick itself) because it is up to the application author to decide how to make the trade-offs that come with using a MonadState over the application state.

> suspendAndResume is a problem for composability, because now you can never compose a suspendAndResume with anything after that, because we want to be able to rely on the result in what follows.

The only way I can see composability changing with that function is to make the event monad a state monad over the application state so that the state changes made by one suspend-and-resume call are available to the next (without explicit passing). Is that what you mean?

> Build in suspendAndResume into EventM, so that the handlers become ...

I'd very much like to avoid changing the top-level event handler type to require the user to return a tuple.  What about this change?
```haskell
-- Old:
-- suspendAndResume :: IO s -> EventM n (Next s)
-- New:
suspendAndResume :: IO s -> EventM n s
```
(With that change, users will be required to explicitly combine that with either `halt` or `continue` and continuing would no longer be implicit. I'd be okay with that change.)

> Don't build in halt into the EventM. This maintains your idea of 'don't call halt in arbitrary places'.
If anyone really needs that still, they can bolt this on by wrapping EventM n s (a, PossiblyHalt) into a newtype that can then have the appropriate Monad instance.

What happens to the API once this is done? Which is to say, how do you envision the API for program termination? (Right now `halt` is *already* not built into the monad at all; it is just encoded in `Next` return value, similarly to what your types suggest above.)

I understand if your project is not open source, and I don't think it's necessary to add me.

NorfairKing:
> because it is up to the application author to decide how to make the trade-offs that come with using a MonadState over the application state.

Fair enough. I'm satisfied with this.

> Is that what you mean?

I don't think so. See below.

> What about this change?
> ```
> -- Old:
> -- suspendAndResume :: IO s -> EventM n (Next s)
> -- New:
> suspendAndResume :: IO s -> EventM n s
> ```

Yes. This would be one of the two parts.
The first part is changing the type of `suspendAndResume` as you propose, and the second part is changing `Next`:

```
-- Old: data Next a = Halt a | SuspendAndResume (IO a) | Continue a
-- New:
data Next a = Halt a | Continue a
```

> I'd very much like to avoid changing the top-level event handler type to require the user to return a tuple. 

This new `Next a` is isomorphic to `(a, PossiblyHalt)`, so I don't care which you prefer.

I hope this answers your question, and you're okay with this change.

Thank you for working with me so patiently on this issue.

jtdaugherty:
> The first part is changing the type of suspendAndResume as you propose, and the second part is changing Next to no longer include this SuspendAndResume constructor

Yeah, that's what I meant about "removing suspendAndResume from Next."

> This new Next a is isomorphic to (a, PossiblyHalt), so I don't care which you prefer (as long as the constructors of Next are exported somewhere).

Yeah, that's the problem: they aren't exported now, and I want it to stay that way.

But I actually want to go back to square one just a bit at this point, because while some changes we've discussed are okay with me, I realized I'm still not convinced that some of the original impetus of this request is actually justified.

The difficulty of composing event handlers that you originally brought up had to do with things like `halt` and `suspendAndResume`, that is, things that had the type
```
thing :: ... -> EventM .. (Next s)
```
where the key barrier to composability was `Next`. The problem that you described was an inability to compose two things that both had this result type. What I realized I am missing is an example of a situation where such composition is *actually necessary*.

So what I need, before we go any further, is an example of a pair of top-level event handlers (i.e. those that return a `Next`) that you need to be able to compose.

jtdaugherty:
I'm closing this, but if you want to continue this discussion, let's re-open #144.

PR #147, opened 2017-10-25
Make events orderable

NorfairKing:
This makes it possible to put events in a map (which is part of my use-case).


jtdaugherty:
Thanks!

PR #148, opened 2017-11-03
Fix a small rst rendering issue

binarin:
Fixed width inline markups runs for longer than intended:

Because according to rst spec:

The end-string must end a text block (end of document or followed by a blank line) or be immediately followed by whitespace or any of ' " . , : ; ! ? - ) ] } / \ or >.


jtdaugherty:
Thanks!

PR #163, opened 2018-03-17
travis-ci: add proper CI build instructions for all major GHC 8.x releases

peti:
There are other options to `make-travis-yml` (https://github.com/haskell-CI/haskell-ci) that might be worthwhile enabling, i.e. `--doctest` or `--hlint`, but I went the conservative route for the beginning.


jtdaugherty:
Thanks! I definitely don't want `--hlint`, and I'm not sure `--doctest` would do much good (there aren't any doctests).

PR #164, opened 2018-03-23
Add missing dependency

trillioneyes:
The build fails for me unless I add this line to build-depends.


jtdaugherty:
What version of GHC are you using?

jtdaugherty:
This dependency needed to be added conditionally. I've fixed it as of d94773c139472bf74088b86ec06095d6cb55185b. Thank you for bringing this to my attention!

PR #165, opened 2018-03-25
Dynamic borders take two

dmwit:
Currently brick borders have a visual oddity: neighboring borders do not connect. You can work around this in simple situations by computing ahead of time where the borders would meet, then manually breaking a single border widget into three pieces (the border before the meet, the single character that connects the two borders, then the border after the meet). This is super annoying and easy to get wrong; it doesn't scale well to complex layouts; and practically speaking it means that people don't do this and just deal with the visual oddity.

The patches below add support for automatically rewriting borders that meet in this way. Users can add a single function call to their top-level widget, and get dynamic borders that react to their neighbors appropriately.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -77,6 +77,8 @@ library
     Brick.Widgets.Edit
     Brick.Widgets.List
     Brick.Widgets.ProgressBar
+    Data.BorderMap
+    Data.IMap
     Data.Text.Markup
   other-modules:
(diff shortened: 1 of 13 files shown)


jtdaugherty:
Thanks so much for all this work!

jtdaugherty:
This is now released in `0.36`. Thanks again for all your work on this!

PR #168, opened 2018-04-07
Add accessor documentation for list widget

ljwall:
I specifically wanted to add `listElements` to the haddock documentation, but have added the other fields for completeness.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -9,7 +9,7 @@
 -- | This module provides a scrollable list type and functions for
 -- manipulating and rendering it.
 module Brick.Widgets.List
-  ( List(listElements, listSelected, listName, listItemHeight)
+  ( List
 
   -- * Constructing a list
   , list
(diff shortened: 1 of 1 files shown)


jtdaugherty:
So you’re aware, we don’t tend to describe these fields as “returning” but I’ll adjust the diff myself. Thank you for the patch!

ljwall:
Ah great, thanks!

I'm a new to Haskell in generally, and Haddock in particular. I wasn't entirely sure what the appropriate style would be in this case where the accessors are exported but the data constructor isn't.

jtdaugherty:
In my experience the Haddock for such accessors is the same either way.

Something else related to this to be aware of is that if the constructor isn't exported, then the accessors *don't appear to be record fields even though they are*, which means that to Haddock readers it won't be evident that expressions like `someList { listSelected = Nothing }` are legal because `listSelected` doesn't appear in the Haddock as part of the `List` record type. If the `List` constructor were exported, then Haddock would render the accessors as part of the type's record syntax as with e.g. http://hackage.haskell.org/package/brick-0.36/docs/Brick-Types.html#t:Widget.

(In this case I am very much okay with that being the case, though, because I don't want users modifying `List` values using those fields directly.)

ljwall:
Thanks for the follow up. I'd not really considered `someList { listSelected = Nothing }` expressions.

Also thanks for having such a thoroughly documented library with many examples --- it is helpful while trying to jump from Learn You a Haskell to actually making working and useful programs!

jtdaugherty:
You're welcome! I'm really glad to hear that the documentation and examples are helpful.

PR #176, opened 2018-05-04
Add renderListI: provides index to the list entry render function

ljwall:
On the list widget, I need to be able to use the entry's index while rendering. That can be done without this change, but I can't see a neat way to do it otherwise.

Maybe this is bloat? I won't be offended if it's declined :-)

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -16,6 +16,7 @@ module Brick.Widgets.List
 
   -- * Rendering a list
   , renderList
+  , renderListWithIndex
 
   -- * Handling events
   , handleListEvent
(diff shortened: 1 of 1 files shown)


src/Brick/Widgets/List.hs
```diff
@@ -16,6 +16,7 @@ module Brick.Widgets.List
 
   -- * Rendering a list
   , renderList
+  , renderListI
```
jtdaugherty: Please rename this to `renderListWithIndex`.

src/Brick/Widgets/List.hs
```diff
@@ -215,7 +230,7 @@ drawListElements foc l drawElem =
 
             drawnElements = flip V.imap es $ \i e ->
                 let isSelected = Just (i + start) == l^.listSelectedL
-                    elemWidget = drawElem isSelected e
+                    elemWidget = drawElem (i + start) isSelected e
```
jtdaugherty: Also please pull `i + start` out as a separate `let` binding such as `idx` since it is now used in multiple places.

ljwall:
Changes made. I used `j` rather than `idx` to avoid ghc warning about it shadowing the above `idx`.

jtdaugherty:
Thank you!

jtdaugherty:
This is now released: https://hackage.haskell.org/package/brick-0.37

PR #177, opened 2018-05-04
Set list selection on move when previous selection was Nothing

pkamenarsky:
Currently, when the selected element in a list is set to `Nothing` moving up or down is impossible. This PR fixes this by setting the selection to the first element if the move direction is positive (i.e. down), otherwise it's set to the last element.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -1,5 +1,5 @@
 name:                brick
-version:             0.36.3
+version:             0.36.4
 synopsis:            A declarative terminal user interface library
 description:
   Write terminal applications painlessly with 'brick'! You write an
--- a/brick.cabal
+++ b/brick.cabal
@@ -1,5 +1,5 @@
 name:                brick
-version:             0.36.4
+version:             0.36.3
 synopsis:            A declarative terminal user interface library
 description:
   Write terminal applications painlessly with 'brick'! You write an
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -306,18 +306,23 @@ listMoveByPages pages theList = do
           in
             return $ listMoveBy nElems theList
 
--- | Move the list selected index by the specified amount, subject to
--- validation.
+-- | Move the list selected index. If the index is `Just x`, adjust by the
+-- specified amount; if it is `Nothing` (i.e. there is no selection) and the
+-- direction is positive, set to `Just 0` (first element), otherwise set to
+-- `Just (length - 1)` (last element). Subject to validation.
 listMoveBy :: Int -> List n e -> List n e
 listMoveBy amt l =
-    let newSel = clamp 0 (V.length (l^.listElementsL) - 1) <$> (amt +) <$> current
+    let current = case l^.listSelectedL of
+          Nothing
+            | amt > 0 -> Just 0
+            | otherwise -> Just (V.length (l^.listElementsL) - 1)
+          current -> current
+        clamp' a b c
+          | a <= b = Just (clamp a b c)
+          | otherwise = Nothing
+        newSel = clamp' 0 (V.length (l^.listElementsL) - 1) =<< (amt +) <$> current
     in l & listSelectedL .~ newSel
-  where
-    current = case l^.listSelectedL of
-      Nothing
-        | amt > 0 -> Just 0
-        | otherwise -> Just (V.length (l^.listElementsL) - 1)
-      current -> current
+
 
 -- | Set the selected index for a list to the specified index, subject
 -- to validation.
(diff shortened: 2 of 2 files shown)


brick.cabal
```diff
@@ -1,5 +1,5 @@
 name:                brick
-version:             0.36.3
+version:             0.36.4
```
jtdaugherty: Please revert this hunk. I only change the version number at release time, which is the only time it's possible to tell what the appropriate change is (by looking at all the commits in the release).

jtdaugherty:
I can see how this change would be helpful, but can you describe how you are getting into a situation where the selected index is `Nothing` even though the list is non-empty?

Essentially this change breaks the semantics of `listMoveBy`, because it is legal to have no selected index at all and this change makes `listMoveBy` change that in a clever way which might not be the caller's intention.

pkamenarsky:
Say I want to display a list of realtime data. The list doesn’t always have the same items - sometimes items may be reordered, disappear, or reappear again.

Let’s say we want to keep the cursor position fixed at the currently selected item, no matter where the item may end up at the next update, so every item must have some sort of unique key and the selected index must be updated to match the key on every update.

Now, if the item disappears, it’d be counterintuitive to just jump to some other position, even less so when the item might reappear a second later. Ideally, the selection would disappear (and reappear at the same item if the item reappears again), but the user must still be able to navigate through the list whenever this happens.

Hope this makes sense. Will remove the version hunk if you decide to go forward with this.

jtdaugherty:
The situation you described makes sense to me. I would imagine, though, that the routine that repopulates the list would also be a good place to decide what happens when the previously-selected element doesn't appear in the new list. In that case, whether to select the first (or last) element is up to your application, so that as soon as the user takes an action (e.g. one that triggers `listMoveBy`), the selected index is already some value that makes sense.

The invariant that the selected index is supposed to help with is: if the list is empty, it is `Nothing`. If it is non-empty, it is *always* `Just`. This is why the `list` constructor always initializes the selected index to `Just 0` for non-empty lists. If you choose to modify the list, it's up to you to decide what the reasonable selection index is. Another reason I would recommend your list population do this is that it avoids the unpleasant UI where your list has been repopulated, but the user takes an action to call `listMoveBy` and suddenly the list selection cursor appears where there was not one before. If you ensure that your selection index is always set appropriately on list repopulation, that will be avoided. Does that make sense?

pkamenarsky:
> I would imagine, though, that the routine that repopulates the list would also be a good place to decide what happens when the previously-selected element doesn't appear in the new list. In that case, whether to select the first (or last) element is up to your application, so that as soon as the user takes an action (e.g. one that triggers listMoveBy), the selected index is already some value that makes sense.

Unfortunately, there's no sensible value other than `Nothing` here. Moving the selection to the first or last or next item all seem counter-intuitive to me, especially when the same item might reappear seconds later.

To give you some more context, I'm working on a drill-down UI with realtime data. The user can select an item and then press a key to see a detailed view (which may also contain another list and so on), then go back or drill down further.

Now imagine that right before the user opens a detailed view the program receives an update *without* the currently selected item, so the selection gets set to (let's say) the first item. At this point the key press event is received and the details view of a completely unrelated item is opened. Worse, the user might not notice their "mistake" right away and continue working under the assumption that they've opened the item which was right under the cursor (which is a very reasonable assumption to make).

If the selection disappeared everything would be fine - the use has an immediate indication that there's no item anymore and drilling down wouldn't work without a selected item. Without doing anything, the item (and the selection) might reappear a bit later. Otherwise they'd have to explicitly select another item (or go back).

> The invariant that the selected index is supposed to help with is: if the list is empty, it is Nothing. If it is non-empty, it is always Just.

Is this just convention or is code depending on this?

In any case, If `brick` were to support this particular use case I think I see two options:

* change above invariant to "if no item is selected the selected index is Nothing; otherwise it's Just"
* or, add a `adjustSelectedIndexIfNothing` flag to `listMoveBy`, propagating it to all calling sites (in particlar `handleListEvevents` and `handleListEventsVi`).

Do you see other alternatives? What do you think?

jtdaugherty:
> Unfortunately, that would mean I'd have to reimplement my own handleListEvent/Vi and everything else that depends on listMoveBy

Okay, that makes sense. Thanks for the discussion! Can you update the Haddock for `listMoveBy` to describe its new behavior? Then I'll merge.

pkamenarsky:
Done. Reverted version, adjusted docs and I also added an additional check to *not* modify the index whenever the list is empty. Let me know if there's anything else.

jtdaugherty:
Thank you!

jtdaugherty:
This is now released: https://hackage.haskell.org/package/brick-0.37

PR #185, opened 2018-09-07
Forms: position cursor between radio brackets

mlang:
Before this change, the cursor would be set to the last column of the line,
which is confusing at least for braille users who rely on cursor position in
terminal program to know where the "focus" is.

### DIFF
--- a/src/Brick/Forms.hs
+++ b/src/Brick/Forms.hs
@@ -78,7 +78,7 @@ module Brick.Forms
   )
 where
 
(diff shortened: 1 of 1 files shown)


mlang:
done

jtdaugherty:
Thank you!

jtdaugherty:
It looks like there's a build error in the latest patch. Can you take a look?

https://travis-ci.org/jtdaugherty/brick/jobs/425815101

PR #187, opened 2018-09-26
Support stm == 2.5.*

osa1:
(in stm-2.5.0.0`newTBQueue` takes a `Natural` argument instead of `Int`)

### DIFF
--- a/src/Brick/BChan.hs
+++ b/src/Brick/BChan.hs
@@ -20,7 +20,7 @@ data BChan a = BChan (TBQueue a)
 -- |Builds and returns a new instance of @BChan@.
 newBChan :: Int   -- ^ maximum number of elements the channel can hold
           -> IO (BChan a)
-newBChan size = atomically $ BChan <$> newTBQueue size
+newBChan size = atomically $ BChan <$> newTBQueue (fromIntegral size)
 
 -- |Writes a value to a @BChan@; blocks if the channel is full.
 writeBChan :: BChan a -> a -> IO ()


jtdaugherty:
For posterity: this change supports previous versions of STM, too.

Thanks for the patch!

jtdaugherty:
This is now released in `0.41.2`.

hvr:
@jtdaugherty In future, please help Hackage Trustees by maintaining accurate upper bounds on `brick`'s dependencies; it's a time & energy consuming process for Trustees trying to reverse engineer upper bounds.

See the current progress of current ongoing efforts restoring build-plans at https://matrix.hackage.haskell.org/package/brick

jtdaugherty:
Thanks, @hvr. I wasn't aware that such a human-driven process existed, and I'm sorry to hear that it can be time-consuming. With that said, I feel a little frustrated with your request because in my experience there has been a long-standing culture of "omit upper bounds, then see what breaks" rather than being more conservative about upper bounds. Both approaches have some merit, and annoy both me and users in different ways. :)

I also want to mention that in this situation, and in others where you have commented on projects of mine, it wasn't entirely clear what the context was for the (unsolicited) comment. I looked around and found this,

https://github.com/haskell-infra/hackage-trustees/blob/master/policy.md

which strikes me as something you might consider linking in your comments unless the context is obvious. Essentially, that policy seems to provide the background for why you're commenting on issues such as this, and without it, I feel a little frustrated by the unsolicited feedback. In this case it is about package maintenance style that creates work for others *that I did not ask for them to do*, so it's hard to know how to respond. The context helps me at least see where the comments are coming from, even if I'm deliberately not making the same trade-offs that the trustees are making.

I'd like to consider tighter upper bounds, but where I come up short is knowing that they're accurate, as you say. Upper bounds will be good for keeping the software building (we're agreed that that's a good thing), but they're going to be overconstrained most of the time, leading to a different user/maintenance burden. What are your thoughts on that?

hvr:
You raise important points and there's a lot to say about this topic, and I'd love to. And in any case I'd strongly argue against "omit upper bounds, then see what breaks"... :-)

...but because it's already late in my timezone for now I only have time to refer you to some material that's been already written and which I feel touches on the topics you've mentioned:

- https://pvp.haskell.org/faq/ (specifically the Qs about upper bounds, and the topic of "knowing what's accurate"; and also why leaving off upper bounds completely undermines the purpose of the PVP & SemVer; and we *need* the PVP in order to make the constraint-solving Hackage/Cabal ecosystem feasible as well as keep the computational cost to validate that the metadata on Hackage is sound within limits what can be realistically accomplished with the computing resources available to haskell.org)

- https://github.com/haskell-infra/hackage-trustees/blob/master/cookbook.md#best-practice-for-managing-meta-data (which tries to elaborate why "omit upper bounds, then see what breaks" doesn't scale - except for some corner cases -- but `brick` isn't one of those; there's also a more rigorous mathematical way to model the cost models taking into account more properties of the dynamics at hand that I haven't had time to write down yet)

- https://github.com/haskell-infra/hackage-trustees/blob/master/revisions-information.md (complements the other materials to explain why revisions are so important)

jtdaugherty:
I'll take a look at those. Thank you!

PR #195, opened 2018-11-13
Prevent vector exception when list is too small to draw correctly

andrevdm:
Vector.slice throws if the index or start is out of bounds.

Adding a `min 0` to num does not work as you can end up with
 idx being out of bounds. Checking for a zero/negative `idx`
 prevents the crash.

See example at https://gist.github.com/andrevdm/83aefb448eed44a9f10764a416bd6b12
  Crashes if the window is made *tiny*, with this fix there is no crash


jtdaugherty:
Thanks!

PR #202, opened 2018-12-05
tests: exit with nonzero status when tests fail

frasertweedale:
The current 'main' has type 'IO Bool' but a return value of 'False'
does not cause the program to exit with nonzero status.  Witnessed
by the following:

  ftweedal% cat foo.hs
  main :: IO Bool
  main = putStrLn "goodbye world" *> pure False

  ftweedal% runhaskell foo.hs
  goodbye world

  ftweedal% echo $?
  0

This causes the cabal test suite to pass, even when there are
errors.  Unless you inspect the QuickCheck output carefully, you
might not notice that a test is failing.  Update 'main' to ensure
that we exit nonzero when quickCheckAll returns 'False'.

### DIFF
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -2,6 +2,9 @@
 {-# LANGUAGE TypeFamilies #-}
 
 import Control.Applicative
+import Data.Bool (bool)
+import System.Exit (exitFailure, exitSuccess)
+
 import Data.IMap (IMap, Run(Run))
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Oh, yikes. Yeah, that should never have been that way. Thank you!

jtdaugherty:
This is now released in `0.43`.

PR #207, opened 2018-12-07
These commits address two discrepancies in list behaviour.

frasertweedale:
`listReplace`
-------------

Haddock says that if given index is not in bounds, 0 is used.
Implementations clamps it to 0..len - 1.  The commit adds a property
test for what the haddock says, and fixes the code.

It may be that the code is correct and the doc needs changing
instead.  Let me know and I will change the test and amend the doc
instead.

`listMoveTo`
-------------

Haddock says that if no selected index, index becomes 0 if delta is
positive and `len - 1` otherwise.  The implementation instead
sets the index to `0` or `len - 1` as stated, *then applies the
delta*.  This behaviour is surprising.  For example, if no selection
and you apply `listMoveDown` or `listMoveUp`, you end up at `1` or
`len - 2` respectively.

Implementations clamps it in `0..len - 1`.  The commit adds a
property test for what the haddock says, and fixes the code.

It may be that haddock is not really correct either, and what you
really want is to move to index `(-1 + amt)` or `(len - amt)`.  So
your clarification on this is desired.

Finally, something that is not addressed by this commit: the
behaviour when `amt == 0`.  Per haddock, this would result in
selected index being `len - 1`.  This is also surprising.  I would
expect either nothing changes, or selection goes to `0`.  If you
clarify this I can also make whatever change is necessary.

Whatever changes happen here I will obviously reflect into my list
generalisation effort, too.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -339,16 +339,14 @@ listMoveByPages pages theList = do
 -- `Just (length - 1)` (last element). Subject to validation.
 listMoveBy :: Int -> List n e -> List n e
 listMoveBy amt l =
-    let current = case l^.listSelectedL of
+    let len = length (l^.listElementsL)
+        newSel = case l^.listSelectedL of
           Nothing
-            | amt > 0 -> Just 0
-            | otherwise -> Just (V.length (l^.listElementsL) - 1)
-          cur -> cur
-        clamp' a b c
-          | a <= b = Just (clamp a b c)
-          | otherwise = Nothing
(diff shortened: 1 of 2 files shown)


jtdaugherty:
Thanks! I think your listReplace change is right. As for listMoveTo, I think your changes are right but I’d also want to play with it to be sure it is unsurprising. The case where amt is zero should leave the list unchanged, I think. The movement behavior when there is no selection is a weird thing to have to worry about in the first place, but the least it could do is be unsurprising. 

Thank you for the patch!

PR #211, opened 2018-12-13
List: add "seek to next" function

frasertweedale:
A common use case is to seek to the next occurrence in the list of
an element matching some predicate (e.g. next unread email).  Add
the 'listSeekToElement' function to satisfy this use case.

'listSeekToElement' is lazy.  Also redefine 'listMoveToElement' in
terms of 'listSeekToElement', making it lazy.  This changes the
constraints on 'listMoveToElement' from (Traversable t) to
(Foldable t, Splittable t), which is reasonable.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -41,6 +41,7 @@ module Brick.Widgets.List
   , listMoveBy
   , listMoveTo
   , listMoveToElement
+  , listFindBy
   , listMoveUp
   , listMoveDown
   , listMoveByPages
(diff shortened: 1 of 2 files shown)


frasertweedale:
@jtdaugherty as usual, please state your preferred function name and I will update the PR :)

src/Brick/Widgets/List.hs
```diff
@@ -545,18 +545,34 @@ splitClamp l i =
...
+-- the beginning.  If no matching element is found, leave the list
+-- unmodified.
+--
+-- /O(n)/.  Only evaluates as much of the container as needed.
+listSeekToElement :: (Foldable t, Splittable t)
+                  => (e -> Bool)
+                  -> GenericList n t e
+                  -> GenericList n t e
+listSeekToElement test l =
+  let
```
jtdaugherty: Please use four-space indents and write `let` bindings as written elsewhere in this module.

jtdaugherty:
Thanks for the patch! I left one comment, but other than that, I think I'd like the new function to be called `listFindBy`.

frasertweedale:
@jtdaugherty thank you; patch updated.

jtdaugherty:
Thank you!

jtdaugherty:
This is released in `0.45`.

PR #212, opened 2018-12-13
Remove reference to Default class from haddock

jhrcek:
Dependency on data-default was removed in https://github.com/jtdaugherty/brick/commit/4a2b90eae7df9b98999c8134524fc61684a9429e#diff-e76c22dbfcd22ba94982edf2323e9024, so removing the outdated mention of the class from the docs as well.


jtdaugherty:
Thank you!

PR #229, opened 2019-05-04
refactor many applications of 'view listElementsL'

frasertweedale:
List / GenericList derives Foldable, which means that in many cases
where 'length (l ^. listElementsL)' was written, we can instead
write 'length l'.  Likewise for 'null'.  Avoid these unnecessary
applications.


jtdaugherty:
Thanks!

PR #230, opened 2019-05-04
add test for List initial selection

frasertweedale:
We do not have a test to confirm that the initial selection for a
list is always 'Just 0' (or 'Nothing' when the list is empty).  Add
a test for this contract.


jtdaugherty:
Thanks!

PR #233, opened 2019-06-11
featured projects username change

callum-oakley:
The redirect still works, I just noticed that this link was to my old username!


jtdaugherty:
Thanks!

PR #236, opened 2019-06-29
add Splittable instance for []

bollu:
I noticed this was missing. I understand that the performance is quite bad, but it's valuable for quick debugging things with `brick`.

Thanks for the lovely library :)

I am curious as to why this instance did not exist already: Is it because it was seen as too expensive / prone to correctness issues (eg. someone feeding an infinite list)?

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -99,6 +99,7 @@ import Brick.Main (lookupViewport)
 import Brick.Widgets.Core
 import Brick.Util (clamp)
 import Brick.AttrMap
+import qualified Data.List as L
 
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for the patch! I'd rather not accept this change, though, precisely because of the performance issue you mentioned. Although it may be suitable for debugging, I'd rather not add it as a built-in instance because it sends the message that it is an intended use of the API (and it isn't). You can certainly provide an orphan instance of your own and use it if you must, though. As for why it wasn't already there, I suspect it was because the original contributor of `Splittable` decided to leave it out due to the O(n) performance of splitting, and because there are other much better data structures already available as instances.

I'm glad you're getting value from the library!

PR #238, opened 2019-07-13
Fixed a few demos

njayinthehouse:
A few of the demos weren't compiling because of a little missing language extension declaration. All fixed!

Thanks for the great package! :)

### DIFF
--- a/programs/BorderDemo.hs
+++ b/programs/BorderDemo.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Main where
 
--- a/programs/BorderDemo.hs
+++ b/programs/BorderDemo.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE CPP #-}
 {-# LANGUAGE OverloadedStrings #-}
 module Main where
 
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Thanks for the patch. What version of Cabal and GHC was this on?

njayinthehouse:
cabal-install version 1.24.0.2, and GHC version 8.0.2. Sorry for not specifying upfront, I'm a bit new to this.

jtdaugherty:
Interesting - it must not be the GHC version, because on Travis, [Brick already builds on 8.0.2](https://travis-ci.org/jtdaugherty/brick/jobs/558098609
).

This language extension being missing has not been a problem on any of the Travis builds before. And I don't see anything in [the Cabal changelog](https://github.com/haskell/cabal/blob/master/cabal-install/changelog) about this behavior changing, either. So I'm confused about why this patch would be necessary.

jtdaugherty:
Can you provide output from the build error that triggered this change?

jtdaugherty:
I suspect the issue is that your version of cabal doesn't support `default-extensions`. The `CPP` extension is already provided in the `cabal` file for all the demos you needed to get building here, which explains why it works on Travis.

jtdaugherty:
What version of the Cabal library is your `cabal` tool built with? I.e., what is the full output of `cabal --version`?

Issue #249, opened 2019-10-30
Size of FocusRing?


romanofski:
I'm currently implementing a way to search and highlight text (see https://github.com/purebred-mua/purebred/issues/279) and use a `FocusRing` to cycle through all matches. I'd love to show a little indication in the status bar similar to what you have in a web browser, e.g.: `2 of 30 matches`. Unfortunately the `FocusRing` API doesn't give me it's size, or access to it's underlying `CList`. 

Would it be possible to give access to either? I think the underlying data structure with an optic might be useful, but for now would be content with access to the overall size of the list.

I'm happy to contribute it as a patch, just asking whether it's something you'd accept :)

jtdaugherty:
I would be happy to accept a patch to add a length function, `focusRingLength :: FocusRing n -> Int`. Thank you!

romanofski:
Hm... now that I have the length/size of the ring I realise that's only half of what I need, since I don't have access to the index of the element in the ring. Arguably since it's a ring there is real index. Might have to come up with an idea on how I solve that, but will have the `focusRingLength` pull request ready tonight.

jtdaugherty:
Ah yeah, that begs the question: does a ring really have a beginning? For your purposes you probably want to keep track of what was "initially" focused, but you could fake that by including a "search result match number" with each match, so your `n` would be a tuple, say.

romanofski:
@jtdaugherty was my thinking really, albeit I thought it to be a bit "hacky". Just sequentially number all results. Might as well be quicker as well rather than looking up the index with every shift in the ring.

PR #250, opened 2019-10-31
Focus: provide a size function

romanofski:
This patch adds a function returning the length of the focus ring. That
is the size returned by the underlying circular list.

Fixes https://github.com/jtdaugherty/brick/issues/249


jtdaugherty:
Thank you!

PR #254, opened 2019-12-01
New initialRenderState function

newhoggy:
Trying to update https://github.com/lspitzner/bricki-reflex to work with `ghc-8.6.5` and need this export since the `RenderState` constructor is internal and hidden.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,69 @@
+{
+  "version": "2.0.0",
+  "tasks": [
+    {
+      "label": "Build",
+      "type": "shell",
+      "command": "bash",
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Thanks for the patch! Unfortunately I'd rather not merge it, because I do not want this function to be part of the public API. There have been a few other instances of people asking for internals to be exposed to support a Reflex API. Here are some historical references in case you want to read about it (and so that I don't have to repeat my reasoning here):
* https://github.com/jtdaugherty/brick/issues/84
* https://github.com/jtdaugherty/brick/pull/85
* https://github.com/jtdaugherty/brick/issues/215

PR #255, opened 2020-01-08
docs: Added screenshots of the demo programs

drola:
When playing with UI libraries it's nice to see how things look like without necessarily downloading and building.


jtdaugherty:
Thank you!

PR #256, opened 2020-01-27
Fix directories not showing up with fileExtensionMatch

Yvee1:
The fileExtensionMatch function is described as:
>A filter that matches any directory regardless of name, or any regular file with the specified extension.

The first part doesn't seem to work, only files with the specified extension show up. This pull request should fix it.

Love the library and documentation by the way, thanks for your work.

### DIFF
--- a/src/Brick/Widgets/FileBrowser.hs
+++ b/src/Brick/Widgets/FileBrowser.hs
@@ -778,5 +778,6 @@ fileTypeMatch tys i = maybe False (`elem` tys) $ fileInfoFileType i
 -- argument of @"xml"@ would match regular files @test.xml@ and
 -- @TEST.XML@ and it will match directories regardless of name.
 fileExtensionMatch :: String -> FileInfo -> Bool
-fileExtensionMatch ext i =
-    ('.' : (toLower <$> ext)) `isSuffixOf` (toLower <$> fileInfoFilename i)
+fileExtensionMatch ext i = case fileInfoFileType i of
+    Just RegularFile -> ('.' : (toLower <$> ext)) `isSuffixOf` (toLower <$> fileInfoFilename i)
+    _ -> True
--- a/src/Brick/Widgets/FileBrowser.hs
+++ b/src/Brick/Widgets/FileBrowser.hs
@@ -779,5 +779,10 @@ fileTypeMatch tys i = maybe False (`elem` tys) $ fileInfoFileType i
 -- @TEST.XML@ and it will match directories regardless of name.
 fileExtensionMatch :: String -> FileInfo -> Bool
 fileExtensionMatch ext i = case fileInfoFileType i of
+    Just Directory -> True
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for the patch! That was indeed a bug.

The new behavior is a little different than what's documented, because it means that even devices, symlinks, etc. will get matched regardless of name. I'm not sure that was the intent of the original function. It probably should be something like:
```
If directory -> True
If regular file -> apply pattern match
otherwise -> False
```
What do you think?

jtdaugherty:
As I think about it more, it should probably also permit symlinks that point to directories. My intention for this matcher was to allow directory traversal while only showing regular files that might be interesting based on extension. For that, the `fileInfoLinkTargetType` holds the target type.

Yvee1:
I agree, so then 
```
If directory -> True
If regular file -> check file extension
If symlink -> apply function recursively
otherwise -> False
```

Yvee1:
Nevermind, you only get the TargetType, not the FileInfo from a symlink.

jtdaugherty:
The symlink case would just be to make it possible to descend into a *directory* if that's the target of a symlink, not to support matching the file target. So:
```
If directory -> True
If regular file -> check file extension
If symlink and the symlink points to a directory -> True
otherwise -> False
```

Yvee1:
Okay, so a symlink to a file matching the given extension is not shown?

jtdaugherty:
> Okay, so a symlink to a file matching the given extension is not shown?

Yes, that's right. My reasoning: the filename to match is the name of the symlink, not its target, since the name of the link is of more significance. But if it's a directory, then the name is unimportant since the main goal for that case is to support traversal of the directory tree.

src/Brick/Widgets/FileBrowser.hs
```diff
@@ -779,5 +779,10 @@ fileTypeMatch tys i = maybe False (`elem` tys) $ fileInfoFileType i
 -- @TEST.XML@ and it will match directories regardless of name.
 fileExtensionMatch :: String -> FileInfo -> Bool
 fileExtensionMatch ext i = case fileInfoFileType i of
+    Just Directory -> True
     Just RegularFile -> ('.' : (toLower <$> ext)) `isSuffixOf` (toLower <$> fileInfoFilename i)
-    _ -> True
+    Just SymbolicLink -> case fileInfoLinkTargetType i of
+        Nothing -> error "impossible"
```
jtdaugherty: I'm not entirely sure this case is impossible. It could be that the link destination could not be read due to permissions issues, so I think this case should result in `False`.

Yvee1: Oh okay I didn't think of that. It should be correct after that second commit I did then.

jtdaugherty: Ah, thanks - I didn't look ahead to the next commit. Thanks!

jtdaugherty:
Thank you!

jtdaugherty:
This change is now released in `0.51`.

Issue #257, opened 2020-01-31
get CircularList from FocusRing


4eUeP:
I want to add `FocusRing` to instance of `ToJSON`:

```hs
instance (ToJSON n) => ToJSON (FocusRing n) where
  toJSON ring = ...
```

but cant get the internal `CircularList` by `FocusRing`, because the lack of exporting data constructor `FocusRing`. Should this be exported?

Or providing a function: `FocusRing n -> [n]` ?

jtdaugherty:
Yes, I'd be okay with adding a new function `focusRingToList :: FocusRing n -> [n]`. That, combined with `focusGetCurrent`, `focusSetCurrent`, and `focusRing`, should provide enough of an API to be useful with `ToJSON` and `FromJSON`. How does that sound?

4eUeP:
Sounds good, I think I can implement this :) 
(will open a pull request later if you'd accept)

jtdaugherty:
Great!

jtdaugherty:
This change is now released in `0.51`.

PR #258, opened 2020-02-01
Focus: add focusRingToList function

4eUeP:
Provide a function which can convert FocusRing to Haskell List.

(N.B. the order of the result may different from the original list given to `Brick.focusRing`)

Fixes: https://github.com/jtdaugherty/brick/issues/257


jtdaugherty:
Thanks!

PR #271, opened 2020-05-10
Support Home and End keys in the editor

romanofski:
This adds handlers to support pressing Home and End keys to jump to the
beginning as well as the end. The keys are typically supported by almost
all text input widgets and now provided by Brick as well.


romanofski:
I've noticed the missing handlers recently and first thought adding it to Purebred. However that seems to be better suited to be generally supported by Brick, unless you think otherwise :)

jtdaugherty:
Thanks for the patch!

PR #272, opened 2020-05-25
RFC export current entry selection function

romanofski:
I'd like to move Purebred to use the Brick file browser. We used to write our own when Brick hasn't had one, but it would be foolish to maintain our own.

One missing function from the public API in Brick I'm missing is the function which handles the default state I can fall back to. That is, setting a new working directory by entering a selected directory or following symlinks. I could replicate this, but I think I would be repeating myself.

Background motivation is, that in Purebred we attach files by selecting them. We use a different word to refer what Brick uses as a selection. We call it "toggling". So a user in Purebred toggles all files to be attached first, then selects the last file which turn Purebred into attaching what is toggled **and** selected (saves an additional key stroke).

I'd be happy to add a little haddock string and update the pull request if you'd be happy to accept the change.

### DIFF
--- a/src/Brick/Widgets/FileBrowser.hs
+++ b/src/Brick/Widgets/FileBrowser.hs
@@ -67,6 +67,7 @@ module Brick.Widgets.FileBrowser
 
   -- * Handling events
   , handleFileBrowserEvent
+  , maybeSelectCurrentEntry
 
   -- * Rendering
   , renderFileBrowser


jtdaugherty:
Thanks for the patch! I'm okay with exporting that function, but before I merge, I want to check one thing. Are you aware that `maybeSelectCurrentEntry` automatically "toggles" any entry for which `fileBrowserSelectable` returns `True`? I ask because your description -- "setting a new working directory by entering a selected directory or following symlinks" -- is right except for that additional behavior, and I wanted to be sure you are aware. The consequence is that if `fileBrowserSelectable` returns `True` for directories, then `maybeSelectCurrentEntry` will *never* descend into directories (which is deliberate). That is easy to avoid, of course, if that isn't your desired behavior.

romanofski:
Sorta? The mail client would only use `selectNonDirectories` for now. If selecting entire directories ever becomes a thing I guess we'll cross that bring when we get to it. For now explicitly toggling/selecting files you need has the advantage that you keep track of what you attach to an e-mail. Yet this would be our use of the filebrowser widget. Others use cases may differ.

I plonk a doc string on the function and update the merge request. Thanks!!

jtdaugherty:
Okay, thanks. I'll add the docstring.

jtdaugherty:
This is now released in `0.54`.

romanofski:
W00hoo!! Thank you so much!

PR #273, opened 2020-06-12
README: Add git-brunch (git checkout utility)

andys8:
Hi, I enjoy working with `brick` and made `git-brunch`. I and some others use it everyday and it's pretty handy. I added it to the project list. I'd be honored if it would be there :)

https://github.com/andys8/git-brunch

Thanks
Andy


jtdaugherty:
Thank you!

PR #277, opened 2020-08-10
Allow building with base-4.14.1.0 (GHC 8.10.2)

RyanGlScott:
While I was in the neighborhood, I also regenerated `.travis.yml`, taking care to automate the `-f demos` flag that was previously hard-coded.


jtdaugherty:
Thank you!

PR #279, opened 2020-08-10
Add window width/height to Context

thomasjm:
Motivation: `vLimitPercent`/`hLimitPercent` are useful functions, but they leave something to be desired when using them inside a large viewport because they use the "available height." In a viewport, this is a large number and not what I want when I'm designing a UI. (In my experiments it was like 999995? Not sure why.)

I want to render a list of items inside a viewport, and have each item's height constrained to 30% of the total window height. After this PR I can write `vLimitPercentWindow` like this:

```haskell
vLimitPercentWindow :: Int -> Widget n -> Widget n
vLimitPercentWindow h' p =
  Widget (hSize p) Fixed $ do
    let h = clamp 0 100 h'
    ctx <- getContext
    let usableHeight = ctx ^. windowHeightL
    let widgetHeight = round (toRational usableHeight * (toRational h / 100))
    withReaderT (availHeightL %~ (min widgetHeight)) $ render $ cropToContext p
```

### DIFF
--- a/src/Brick/Types.hs
+++ b/src/Brick/Types.hs
@@ -33,10 +33,12 @@ module Brick.Types
   , getContext
 
   -- ** The rendering context
-  , Context(ctxAttrName, availWidth, availHeight, ctxBorderStyle, ctxAttrMap, ctxDynBorders)
+  , Context(ctxAttrName, availWidth, availHeight, windowWidth, windowHeight, ctxBorderStyle, ctxAttrMap, ctxDynBorders)
   , attrL
   , availWidthL
   , availHeightL
+  , windowWidthL
+  , windowHeightL
   , ctxAttrMapL
   , ctxAttrNameL
   , ctxBorderStyleL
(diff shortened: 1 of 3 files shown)


jtdaugherty:
Thanks, I'm happy to merge this change because I can see how it would help you do what you want.

As for your question about viewport height, the answer is that when viewport contents are rendered they are rendered with a synthetic height of 100,000 rows rather than the height of the enclosing viewport. This is because we need to provide enough height that the contents are effectively unlimited (in height, say) but not so high that we get an infinitely-high (or wide) widget which would naturally exhaust all memory and never terminate the renderer. Consequently, viewport contents are expected not to depend on the viewport size to render themselves, and in practice this ends up working out because they usually rely on the width only (in a vertical viewport, say) or on their own computations. However, I can see how this change helps you provide a useful hint to the viewport contents' rendering function.

thomasjm:
Got it, thanks for merging!

thomasjm:
@jtdaugherty any chance you could cut a release containing this change? I've love to start using it in an open-source library.

jtdaugherty:
Done!

http://hackage.haskell.org/package/brick-0.56

thomasjm:
Thanks!

PR #286, opened 2020-08-31
Add a lens into form state

sordina:
I've found that I've had to write a fair bit of boilerplate to rebuild my form after extracting the state for modifications. It would be convenient if the Form fields were also lenses, but especially the state field.

This seems to make sense to me as well since the state itself often requires lenses to be used in form fields, so it's nicely composable if a lens exists for the state.

Let me know if this doesn't make sense for some reason! I've only been using Brick for a brief amount of time.

### DIFF
--- a/src/Brick/Forms.hs
+++ b/src/Brick/Forms.hs
@@ -50,6 +50,7 @@ module Brick.Forms
   , newForm
   , formFocus
   , formState
+  , formStateLens
   , handleFormEvent
   , renderForm
   , renderFormFieldState
@@ -207,6 +208,14 @@ data Form s e n =
          , formConcatAll :: [Widget n] -> Widget n
          -- ^ Concatenation function for this form's field renderings.
          }
+
+-- | A lens into the formState of a form.
+--
+-- This could be useful if you want to preseve focus properties, etc.
+-- of your form but need to modify the underlying application state.
+
+formStateLens :: Lens' (Form s e n) s
+formStateLens = lens formState (\f s -> f { formState = s })
 
 -- | Compose a new rendering augmentation function with the one in the
 -- form field collection. For example, we might put a label on the left


jtdaugherty:
Thanks for the patch!

The reason this doesn't already exist is because if the form state is allowed to change via a lens then it can become out of sync with the contents of the form fields. This is because the form state itself is distinct from the form field states that the form keeps track of for each input (e.g. editor contents, radio button selection, etc.). Those field states need to be built from the initial form state and then as the form field states change, those states need to be written back to the form state. Thus the form state and field states need to be carefully kept in sync.

Right now it isn't possible to do anything but just build a new `Form` from your state. Since the `Form` abstraction is really only intended to handle the simple case where you manage a set of values using a form without mutating it yourself, if that use case doesn't match yours then it may very well be necessary to implement your own form abstraction.

jtdaugherty:
By the way, I am working on an alternative approach to address this problem. Once I've confirmed whether it works, I'll update here.

sordina:
Hi @jtdaugherty thanks for the response!

Yes, I realised last night the exact issue with my idea that you brought up. I look forward to seeing your solution!

In the mean time, could I construct a null form (besides focus info) and have my show function create the form? I'm trying to follow an elm like pattern and having the form object "persisting" is quite counter to that intuition.

jtdaugherty:
If you do as you describe, that will make some form operations impossible (such as editing text fields) since you will not be able to persist that state across form creations (cursor position, say).

As I see it, at the end of the day, avoiding state is not inherently virtuous. Instead, it is best to *constrain* how it is used so it can be used responsibly. In this case there is the application state and there is *also* interface state (such as editor cursor position). You can't get around having to track that somehow.

sordina:
Makes sense to me. Thanks again for your thoughts!

jtdaugherty:
@sordina Please check out b695d79c256363178c6b86a18f2654397df5a208 and see if that works for you.

jtdaugherty:
(And see also 269fb1b205a3c89c40db255632c1e862cb24ae7a)

sordina:
@jtdaugherty awesome. I'll check it out!

PR #292, opened 2020-11-15
Update constraint on base for build with ghc-9.0

mimi1vx:
Brick build works withouts problems with upcoming ghc-9.0.1+ and base-4.15.0.0


jtdaugherty:
Thank you!

PR #294, opened 2020-12-05
FileBrowser: Put cursor where selected file is printed

mlang:


jtdaugherty:
Thank you!

mlang:
I believe there might be a slight oversight in my patch.
I guess the condition should be `foc && listSel`, and not just `listSel`?

If you agree I can follow up with a correction.

jtdaugherty:
Ah yes, I think you’re right.

PR #296, opened 2020-12-05
Only position cursor when FileBrowser is actually focused

mlang:


jtdaugherty:
This change has been released in `0.58`.

PR #299, opened 2020-12-17
Brick.Widget.List: add listCycleFindBy

iyefrat:
This function is like `listFindBy`, but it loops back from the top if it didn't find anything, up to completing one full cycle. Personally this is what I expected `listFindBy` to do, but I assumed it would be better to add this behaviour in a different function.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -50,6 +50,7 @@ module Brick.Widgets.List
   , listMoveTo
   , listMoveToElement
   , listFindBy
+  , listCycleFindBy
   , listMoveUp
   , listMoveDown
   , listMoveByPages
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for the patch! Reading the docstring for `listFindBy`, I think it should do what this is doing. The first sentence already suggests this: "Set the selected index to the next element matching the predicate." I think it would be better to just fix `listFindBy` to do search from the beginning if a match isn't found starting at the selection position. I went ahead and did that in bcebc3a9bb1ef742e4ecdd482a6188c2f53b1c85.

iyefrat:
Neat. I like your implementation better as well.

jtdaugherty:
This change has been released in `0.58`.

PR #300, opened 2020-12-18
Suggestion: add Filterable (and Witherable) instances for GenericList

iyefrat:
This entails adding [witherable-class](https://hackage.haskell.org/package/witherable-class) as a dependency.

If you are unaware, this is a package that introduces the typeclass `Filterable` that polymorphises the function `filter`, and introduces the function `wither`, which gives you a filtering `traverse`. `witherable-class` is a dependency of its heavier counterpart `witherable`, which is why it's relatively small (although it does import `hashable`).

I suggest this because writing `W.filter f someList` is much more readable to me than for example `someList & listElementsL %~ V.Filter f`, and it lets us define filters on `GenericList n f e` polymorphically on `f`.

I can of course just implement this on my end using an orphan instance, but I thought I should bring it to your attention in case you thought it would be useful to add to `brick`.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -135,7 +135,8 @@ library
                        deepseq >= 1.3 && < 1.5,
                        unix,
                        bytestring,
-                       word-wrap >= 0.2
+                       word-wrap >= 0.2,
+                       witherable-class
   if impl(ghc < 8.0)
     build-depends:     semigroups
 
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -73,7 +73,7 @@ module Brick.Widgets.List
   )
 where
 
-import Prelude hiding (reverse, splitAt)
+import Prelude hiding (reverse, splitAt, filter)
 
 import Control.Applicative ((<|>))
 #if !MIN_VERSION_base(4,8,0)
@@ -97,6 +97,7 @@ import qualified Data.Sequence as Seq
 import Graphics.Vty (Event(..), Key(..), Modifier(..))
 import qualified Data.Vector as V
 import GHC.Generics (Generic)
+import qualified Data.Witherable.Class as W
 
 import Brick.Types
 import Brick.Main (lookupViewport)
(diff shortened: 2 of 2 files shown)


jtdaugherty:
Thanks for the patch! I think I'd be okay with adding this, but before I do, the instance needs to account for changes in the selection due to the filter result. Right now, if an item is selected, then after the filter the selection may be wrong in a couple different ways:
* If the filter removed all elements, the selection should be set to `Nothing`.
* If the filter resulted in changing the position of the selected element, the selection should be updated.
* If the filter resulted in removing the selected element, the selection should *probably* be set to the preceding item, if any.

(And those behaviors will need to be documented somewhere prominent.)

iyefrat:
> If the filter removed all elements, the selection should be set to Nothing.

I agree, I'll add this.

> If the filter resulted in changing the position of the selected element, the selection should be updated.
> If the filter resulted in removing the selected element, the selection should probably be set to the preceding item, if any.

I'm a bit conflicted on this. The reason I implemented a filter function in the first place is because I have a mechanism to highlight and delete items from a list. In my mind, the intuitive outcome after deleting the items if for the cursor to stay in the same place visually, while the list changes under it, which is the current behaviour. However this is subjective and I can see the argument for the intuitive outcome being the cursor staying on the same item (or as close as it can manage if the item is deleted, which is always true in my use case).

What do you think?

jtdaugherty:
I can see it both ways, which is actually the reason why I have any reservations about this patch (and I do). If this patch is merged, then I want it to do *the least surprising thing* when the user has no particular intention. That is still subjective, but to me that means preserving the selection not in terms of position (which I suspect is *rarely* the semantically important thing about the selection) but the item being selected.

Although I can understand the utility of this patch, it does break the `List` abstraction in that it exposes the underlying sequence for manipulation in more ways than are already provided. If we're going to do that, then we do need to think about maintaining sane behavior of the selection. I would rather avoid this patch if we can't come up with sane behavior, because then at least we force the user (yourself, in this case) to make a policy decision on how the list should be rebuilt after a filter just by using the existing API to build a new list and then select the desired element.

iyefrat:
I've thought about this, and I agree with you that keeping the cursor on the same item (or close to it if the selected item gets deleted) is a better default. However, due to the fact that there is no one "obvious" way to implement this, I find myself questioning a bit if this is not in fact best left in an orphan instance. Even if we implement the filter to keep the selection on the same semantic item, if the selected item gets deleted, the choice of whether or not the selection moves to the item above or below is entirely arbitrary (I lean towards below).

What do you think, is it a reasonable enough default to be worth including, or is this feature best left in an orphan instance?

jtdaugherty:
Since there is this extra policy decision that we have been talking about, it sounds like an orphan instance might be best. Type classes like this are really good when there is one and only one way such operations should be implemented for a structure, but they are less useful when there are multiple valid ways to do so (another type class that suffers from this problem is quickcheck's `Arbitrary` class).

Thanks for discussing it with me!

PR #304, opened 2020-12-27
cbookview: A TUI for exploring (polyglot) chess opening book files

mlang:


jtdaugherty:
Nice! Would you be willing to include a screenshot of the TUI in your Readme?

mlang:
Me is blind and a plain terminal user.  Contrinbutions welcome :)

jtdaugherty:
The reason I ask you is because you're in a better position to know what is worth including in a screenshot, having designed the TUI and understanding what the application does better than I do. I don't think I'm in a position to know what a good screenshot could include. I like to link to Brick-based projects in the README, and screenshots are one way to help people understand what they accomplished.

Still, if you don't think adding one makes sense, that's okay - just let me know.

mlang:
I am blind.  I can't do a classical screenshot without sighted help.

mlang:
I just added an ASCII "screenshot" to the README which should
give the user an idea what `cbookview` looks like.

BTW: `cbookview` is almost an example program for `brick`.
It couldn't be much simpler.  It draws all its power
from the `chessIO` and `rosezipper` packages.  The app state
is basically just a `TreePos` from `rosezipper` and thats it.  The time
taken from the initial idea to the first working implementation with
`brick` was also pretty jaw dropping for me.  It felt extremely short
given that I always wanted such a tool and it appeared to just emerge
out of using the right data structures.

jtdaugherty:
Thanks for the explanation, that makes sense. The ASCII you posted was pretty much the sort of thing I had in mind; I figured if you were wanting to add this to the Brick application list in my README, then there must be *something* you wanted to show. 😊 

I'm also glad to hear that it sounds like you had a good experience putting together the TUI!

PR #305, opened 2021-01-07
be careful about rewriting empty images

dmwit:
The following short program crashes when the terminal is 21 rows tall:

```haskell
import Brick
import Brick.Widgets.Border

main :: IO ()
main = simpleMain . joinBorders . vBox $ [str (replicate 20 '\n') :: Widget (), border emptyWidget, hBorder]
```

This happens because the border rewriting code assumes it is rewriting a vty `Image` that actually has content, but the bottom part of the border has already been cropped to height 0. This pull request has a small patch that fixes this by skipping the rewrite if the image has no height (or width, if we are rewriting for an `hBox`).


jtdaugherty:
Thank you!

PR #307, opened 2021-02-05
Include clickable elements in the rendering cache.

phsmenon:
Mouse events for `Widget` were being received inconsistently if/when they are cached. This appears to be because a `Widget`, when rendered update the list of clickable elements that `brick` should track for mouse activity. However, when a `Widget` is cached, this information is lost.

This commit updates the rendering cache to also include the clickable elements exposed by a `Widget` and update `brick`'s tracking list appropriately when utilizing the cache.

### DIFF
--- a/src/Brick/Widgets/Core.hs
+++ b/src/Brick/Widgets/Core.hs
@@ -962,6 +962,11 @@ vRelease p =
 -- use the rendered version from the cache. If not, render the specified
 -- widget and update the cache with the result.
 --
+-- To ensure that mouse events are emitted correctly for cached widgets, 
+-- in addition to the rendered widget, we also cache (the names of) 
+-- any clickable extents that were rendered and restore that when utilizing
+-- the cache.
+--
 -- See also 'invalidateCacheEntry'.
 cached :: (Ord n) => n -> Widget n -> Widget n
 cached n w =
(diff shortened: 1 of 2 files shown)


src/Brick/Widgets/Core.hs
```diff
@@ -968,18 +968,27 @@ cached n w =
...
         result <- cacheLookup n
         case result of
-            Just prevResult -> return prevResult
+            Just (clickables, prevResult) -> do
+                clickableNamesL %= (clickables ++)
+                return prevResult
             Nothing  -> do
                 wResult <- render w
-                cacheUpdate n wResult
+                clickables <- renderedClickables wResult
```
jtdaugherty, line 976: I believe the list comprehension in `renderedClickables` is redundant, and this can be reduced to just `clickables <- use allClickablesL`.

jtdaugherty: Oh, I see why you did this now. You are limiting the clickable names to just the clickable names whose extents were in the rendered result. Please disregard my comment!

jtdaugherty: (Could you add a Haskell comment to this code explaining that nuance?)

jtdaugherty:
I left just the one bit of commentary above; otherwise this looks great to me. Thank you!

phsmenon:
I have added comments to the code about the updates and pushed it. Hope they look ok?

jtdaugherty:
Yes, those comments are helpful. Thank you!

PR #309, opened 2021-03-11
Fix condition for base version.

mimi1vx:
Unfortunately `base <=4.15` is false for `base-4.15.0.0` so fix condition to `base < 4.16`


jtdaugherty:
Thanks!

PR #310, opened 2021-03-15
Use of infix functions over lambda expressions

KommuSoft:


jtdaugherty:
Similar to the other infix change, I don't think this change improves readability. Thank you for the patch, though!

PR #313, opened 2021-03-15
Use sectioning over flip

KommuSoft:


jtdaugherty:
I'm torn about this change. While I often find `flip` hard to think about, I find sections even harder to read when the functions in question are not really designed to read as infix. I think I'm going to pass on this one. Thank you for the patch, though!

PR #317, opened 2021-04-08
Adds editShowableFieldWithValidate

benjaminselfridge:
This function for Brick.Forms allows the user to specify an additional function
to validate input in addition to `readMaybe`, as a slightly more flexible
version of `editShowableField`.


jtdaugherty:
Thanks!

PR #327, opened 2021-06-29
Simplify

mlang:
foo <- bar; return $ f foo = fmap f bar
(& foo) = foo

### DIFF
--- a/src/Brick/Widgets/Core.hs
+++ b/src/Brick/Widgets/Core.hs
@@ -153,12 +153,14 @@ class Named a n where
 -- | When rendering the specified widget, use the specified border style
 -- for any border rendering.
 withBorderStyle :: BorderStyle -> Widget n -> Widget n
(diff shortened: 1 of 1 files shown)


src/Brick/Widgets/Core.hs
```diff
@@ -992,32 +995,28 @@ cropBottomTo rows p =
...
 -- | When rendering the specified widget, also register a cursor
 -- positioning request using the specified name and location.
 -- The cursor will only be positioned but not made visible.
 putCursor :: n -> Location -> Widget n -> Widget n
-putCursor n cloc p =
-    Widget (hSize p) (vSize p) $ do
-      result <- render p
-      return $ result & cursorsL %~ (CursorLocation cloc (Just n) False:)
+putCursor n cloc p = Widget (hSize p) (vSize p) $
+    fmap (cursorsL %~ (CursorLocation cloc (Just n) False:)) (render p)
```
jtdaugherty, line 1006: I prefer infix `fmap` (`<$>`) when possible. I'll merge and update this case, but I just wanted you to know for future patches. Thank you!

PR #331, opened 2021-07-20
Edit: Bind more Emacs-compatible keys

mlang:
M-b and M-f to navigate by word
C-b and C-f for consistency
M-d to delete word under cursor
C-t to transpose previous char with current char

This replaces the Monoid constraint on handleEditorEvent with GenericTextZipper

### DIFF
--- a/src/Brick/Widgets/Edit.hs
+++ b/src/Brick/Widgets/Edit.hs
@@ -52,6 +52,7 @@ import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import qualified Data.Text.Zipper as Z hiding ( textZipper )
 import qualified Data.Text.Zipper.Generic as Z
+import qualified Data.Text.Zipper.Generic.Words as Z
 
 import Brick.Types
 import Brick.Widgets.Core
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thanks for this patch. I'm going to merge it, but I do want to mention that I see a risk in adding more Emacs-style bindings to the event handler. The more it supports, the more it may inadvertently set the expectation that Emacs bindings are, in general, going to be supported. That is not an expectation I want to set, and I don't want to get into a slippery slope situation where adding some means users then feel disappointed by the lack of other Emacs-style bindings (and, as we both probably know, those are basically limitless). So while these bindings make sense to me as basic functionality, I am just wanting to keep an eye on whether the set of Emacs bindings grows too large at some point (and I don't know how to quantify that). If that does happen, it may make sense to factor out the event handler into a new package that makes more of an explicit attempt to provide Emacs-style bindings.

mlang:
I read the note about Editor not being a base widget for a full editor,
and I get the point.  I will not actively add anything to TextZipper, so at
least from my side there is no real risk of bloat.  However, these gems
were just too nice to not bind by default.  Besides, I only mentioned
"Emacs-compatible" in the commit log since the current bindings
are clearly Emacs-derived (C-k, C-d, C-a, C-e).  Thats nice insofar that
it is not hard to come up with bindings for well-known behaviour.
Better not invent the wheel anew...

jtdaugherty:
Actually, I would genuinely appreciate contributions to `TextZipper`. And I appreciate you adding these new bindings, particularly if the underlying functionality already existed in the zipper. I just want to make it clear that the Brick editor is not necessarily trying to be Emacs, which you understand. And if someone wanted more Emacs-like functionality, I could imagine just moving the event handler to another package where someone can go wild adding that functionality while the base library's event handler can provide "basic" editing functionality. That, at least, would give the best of both worlds: brick has good batteries included, but that doesn't put up a barrier to more advanced functionality.

jtdaugherty:
This change is now released in 0.64. Thanks again!

PR #344, opened 2021-11-09
Mention table widget in README.md

NiekPas:
Feel free to disagree, but I think the table widget is cool enough to deserve a frontpage mention :)


jtdaugherty:
Thanks!

PR #346, opened 2021-11-18
add hledger-ui, hledger-iadd to the projects list

simonmichael:
I'm not sure how it's ordered, so I just added these at the bottom. But hledger-ui has been using brick since 2015, and hledger-iadd since 2016.


jtdaugherty:
Awesome, thanks!

PR #358, opened 2022-01-28
Use guards instead of if-else

mlang:


jtdaugherty:
Thanks for submitting these patches. In general, I'm more open to changes that change or add functionality rather than just change code style. I'm unlikely to write code in the style of these particular patches, which means that if I came across this code I'd be likely to rewrite it. That seems like unnecessary churn, so I'm going to decline this PR.

PR #373, opened 2022-06-19
make empty BorderMaps even emptier (resolves issue 370)

dmwit:
The BorderMap module provides a notion of an empty map. Previously, this
was accidentally conflating two kinds of emptiness: an map that doesn't
track points, and a map that tracks a point (or multiple points) but
doesn't associate it with anything. This meant that the empty map was
not a unit for the union operation, which seems like a natural
expectation.

This has been resolved by revising the empty map to be the one that
tracks no points. Callers that expected that behavior were left alone,
while callers that expected the other behavior were fixed to use
`emptyCoordinates`.


jtdaugherty:
Thank you!

PR #374, opened 2022-07-01
(RFC/proposal/WIP) renderWidget function

valyagolev:
This is a work-in-progress Pull Request that I'm making to ask for your opinion on the proposed changes.

This adds a new function, `renderWidget`, that renders a list of widgets into a `V.Picture`. This can be useful with `ghcid` (so one can enjoy hot-reloading their code while working on the UI), and also opens up possibilities for "golden" testing (that would compare the rendered outputs with the golden output).

One way to go about golden testing would be to use `vty`'s `mockTerminal` – that one doesn't output anything pretty, however it can be used to compare the outputs and, should comparison fail, render it in the real terminal. I didn't implement this completely, as I wanted to ask for your input before endeavouring this. If this sounds like a good idea, I could add a couple of simple golden tests as a starting point. They might prove useful for the continued development of the library.

To try using this thing in `ghcid`, the functions defined in `tests/Render.hs` are sufficient.

Unfortunately, there's no way to get this functionality without changing the library somehow, that's why I have to propose this change. An alternative route would be to expose `RenderState` completely, thus making `renderFinal` usable from the outside code. I believe the proposed change is a better approach and can be quite useful even for the library itself.

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -511,10 +511,11 @@ test-suite brick-tests
   ghc-options:         -Wall -Wcompat -Wno-orphans -O2
   default-language:    Haskell2010
   main-is:             Main.hs
-  other-modules:       List
+  other-modules:       List Render
   build-depends:       base <=5,
                        brick,
                        containers,
                        microlens,
                        vector,
+                       vty,
                        QuickCheck
--- a/src/Brick/Main.hs
+++ b/src/Brick/Main.hs
@@ -47,6 +47,7 @@ module Brick.Main
   , renderFinal
   , getRenderState
   , resetRenderState
+  , renderWidget
   )
 where
 
--- a/src/Brick/Widgets/Internal.hs
+++ b/src/Brick/Widgets/Internal.hs
@@ -4,6 +4,7 @@ module Brick.Widgets.Internal
   , cropToContext
   , cropResultToContext
   , renderDynBorder
+  , renderWidget
   )
 where
 
@@ -13,6 +14,7 @@ import Control.Monad (forM_)
 import Control.Monad.Trans.State.Lazy
 import Control.Monad.Trans.Reader
 import qualified Data.Map as M
+import qualified Data.Set as S
 import Data.Maybe (catMaybes)
 import qualified Graphics.Vty as V
 
@@ -154,3 +156,27 @@ renderDynBorder db = V.char (dbAttr db) $ getBorderChar $ dbStyle db
             Edges True  True  False True  -> bsIntersectL
             Edges True  True  True  False -> bsIntersectR
             Edges True  True  True  True  -> bsIntersectFull
+
+-- | This is a debug-only function to easily render a list of Widgets
+-- as a 'V.Picture'. You can use it with ghcid, or even for golden testing
+-- with 'Graphics.Vty.Output.Mock.mockTerminal'.
+renderWidget :: (Ord n) 
+            => AttrMap 
+            -> [Widget n] 
+            -> V.DisplayRegion 
+            -> ([CursorLocation n] -> Maybe (CursorLocation n)) 
+            -> (V.Picture, Maybe (CursorLocation n), [Extent n])
+renderWidget aMap layerRenders (w, h) chooseCursor =
+  (pic, mloc, exts)
+  where
+    initialRS =
+      RS
+        { viewportMap = M.empty,
+          rsScrollRequests = [],
+          observedNames = S.empty,
+          renderCache = mempty,
+          clickableNames = [],
+          requestedVisibleNames_ = S.empty,
+          reportedExtents = mempty
+        }
+    (_, pic, mloc, exts) = renderFinal aMap layerRenders (w, h) chooseCursor initialRS
--- a/src/Brick/Widgets/Internal.hs
+++ b/src/Brick/Widgets/Internal.hs
@@ -157,17 +157,40 @@ renderDynBorder db = V.char (dbAttr db) $ getBorderChar $ dbStyle db
             Edges True  True  True  False -> bsIntersectR
             Edges True  True  True  True  -> bsIntersectFull
 
--- | This is a debug-only function to easily render a list of Widgets
--- as a 'V.Picture'. You can use it with ghcid, or even for golden testing
--- with 'Graphics.Vty.Output.Mock.mockTerminal'.
+-- | This functions renders a list of 'Widget's as a 'V.Picture',
+-- bypassing the interactive components. It can be helpful in case
+-- you only need the output; still in general it's preferrable to use
+-- the main, interactive API. As of now, and before there's a proven use and need,
+-- this function is considered an exposed internal mechanism,
+-- doesn't promise to address all the corner cases and must be used with caution.
+--
+-- Consult [vty docs](https://hackage.haskell.org/package/vty-5.35.1/docs/Graphics-Vty-Output.html)
+-- on how to output the resulting 'V.Picture'.
+--
+-- An example function that renders a couple of Widgets and outputs the result on the screen:
+--
+-- @
+--      import qualified Graphics.Vty as V
+--
+--      renderDisplay :: Ord n => [Widget n] -> IO ()
+--      renderDisplay ws = do
+--        outp <- V.outputForConfig V.defaultConfig
+--        ctx <- V.displayContext outp region
+--        V.outputPicture ctx (renderToPicture ws)
+--
+--      myRender :: IO ()
+--      myRender = do
+--        renderDisplay @() [str "Why" <=> hBorder <=> str "Not?"]
+--        putStrLn "" -- this empty line makes sure the cursor assumes
+--                    -- the correct position after the output
+-- @
+--
+-- As a suggestion for usage, you can use it with `ghcid` to quickly iterate on how a Widget looks.
 renderWidget :: (Ord n) 
-            => AttrMap 
-            -> [Widget n] 
+            => [Widget n] 
             -> V.DisplayRegion 
-            -> ([CursorLocation n] -> Maybe (CursorLocation n)) 
-            -> (V.Picture, Maybe (CursorLocation n), [Extent n])
-renderWidget aMap layerRenders (w, h) chooseCursor =
-  (pic, mloc, exts)
+            -> V.Picture
+renderWidget layerRenders (w, h) = pic
   where
     initialRS =
       RS
@@ -179,4 +202,4 @@ renderWidget aMap layerRenders (w, h) chooseCursor =
           requestedVisibleNames_ = S.empty,
           reportedExtents = mempty
         }
-    (_, pic, mloc, exts) = renderFinal aMap layerRenders (w, h) chooseCursor initialRS
+    (_renderState, pic, _mloc, _exts) = renderFinal (attrMap V.defAttr []) layerRenders (w, h) (const Nothing) initialRS
--- a/tests/Main.hs
+++ b/tests/Main.hs
@@ -13,6 +13,7 @@ import qualified Data.IMap as IMap
 import qualified Data.IntMap as IntMap
 
 import qualified List
+import qualified Render
 
 instance Arbitrary v => Arbitrary (Run v) where
     arbitrary = liftA2 (\(Positive n) -> Run n) arbitrary arbitrary
@@ -111,5 +112,5 @@ return []
 
 main :: IO ()
 main =
-  (all id <$> sequenceA [$quickCheckAll, List.main])
+  (all id <$> sequenceA [$quickCheckAll, List.main, Render.main])
   >>= bool exitFailure exitSuccess
--- a/tests/Render.hs
+++ b/tests/Render.hs
@@ -1,0 +1,32 @@
+{-# LANGUAGE TypeApplications #-}
+module Render
+  ( main,
+  )
+
+where
+
+import Brick
+import qualified Graphics.Vty as V
+import Brick.Widgets.Border (hBorder)
+
+region :: V.DisplayRegion
+region = (30, 10)
+
+renderToPicture :: Ord n => [Widget n] -> V.Picture
+renderToPicture ws = let (p, _, _) = renderWidget (attrMap V.defAttr []) ws region (const Nothing) in p
+
+renderDisplay :: Ord n => [Widget n] -> IO ()
+renderDisplay ws = do
+  outp <- V.outputForConfig V.defaultConfig
+  ctx <- V.displayContext outp region
+  V.outputPicture ctx (renderToPicture ws)
+
+ui :: Widget ()
+ui = str "Hello, world!"
+
+main :: IO Bool
+main =
+  do 
+      renderDisplay [ui]
+      renderDisplay @() [str "why" <=> hBorder <=> str "not"]
+      return True
--- a/tests/Render.hs
+++ b/tests/Render.hs
@@ -8,6 +8,7 @@ where
 import Brick
 import qualified Graphics.Vty as V
 import Brick.Widgets.Border (hBorder)
+import Control.Exception (try)
 
 region :: V.DisplayRegion
 region = (30, 10)
@@ -28,8 +29,11 @@ renderedMyWidget = "Picture ?? [VertJoin {partTop = VertJoin {partTop = HorizTex
 main :: IO Bool
 main =
   do 
-      renderDisplay [myWidget]
-      
+      result <- try (renderDisplay [myWidget]) :: IO (Either V.VtyConfigurationError ())
+      case result of
+        Left _ -> putStrLn "Terminal is not available"
+        Right () -> return ()
+
       print $ show $ renderWidget [myWidget] region
 
       return $
--- a/tests/Render.hs
+++ b/tests/Render.hs
@@ -12,21 +12,25 @@ import Brick.Widgets.Border (hBorder)
 region :: V.DisplayRegion
 region = (30, 10)
 
-renderToPicture :: Ord n => [Widget n] -> V.Picture
-renderToPicture ws = let (p, _, _) = renderWidget (attrMap V.defAttr []) ws region (const Nothing) in p
-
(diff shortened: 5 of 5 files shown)


jtdaugherty:
Thanks for opening this.

> This can be useful with ghcid (so one can enjoy hot-reloading their code while working on the UI)

I have used `ghcid` before, but based on my understanding of what it does I don't quite understand this bit. Can you say a bit more about how this is useful in the context of `ghcid`? Is there some issue with hot-reloading that is resolved by having this function?

> One way to go about golden testing would be to use vty's mockTerminal – that one doesn't output anything pretty, however it can be used to compare the outputs and, should comparison fail, render it in the real terminal. I didn't implement this completely, as I wanted to ask for your input before endeavouring this. If this sounds like a good idea, I could add a couple of simple golden tests as a starting point.

In this comment, are you referring to adding golden tests to `brick` itself?

As far as the approach taken in the patch, I prefer this over exposing any internals. Internals, by definition, are easy to use incorrectly and are not designed for outside consumption. So it makes sense to me that we'd need a new function that bundled up the internals and exposed the operation at a higher level.

valyagolev:
> I have used ghcid before, but based on my understanding of what it does I don't quite understand this bit. Can you say a bit more about how this is useful in the context of ghcid? Is there some issue with hot-reloading that is resolved by having this function?

Hot-reloading doesn't work too well with interactive apps. I don't think it'd be easy to run a brick app interactively in `ghcid` – it doesn't really seem to try to forward the keyboard events. And even if it'd be possible, I'm not sure about the convenience of this approach, as a naively implemented reload would reset the state.

I don't really know how people typically develop brick apps, perhaps they just run them again every time. In some cases, especially when working on UI, I think `ghcid` could provide a very easy and tight feedback loop if one only renders the widget (or whole view) they are working on. If there's other ways to achieve this kind of fast feedback loop, I'd be happy if you'd suggest something. Otherwise I think this particular solution can cover a lot of cases.

> In this comment, are you referring to adding golden tests to brick itself?

This changes enables writing golden tests (for any app), and thus can also allow writing the golden tests for brick. One benefit of writing a few such tests for brick is that they can be used as an example of writing golden tests for other apps.

> As far as the approach taken in the patch, I prefer this over exposing any internals. Internals, by definition, are easy to use incorrectly and are not designed for outside consumption. So it makes sense to me that we'd need a new function that bundled up the internals and exposed the operation at a higher level.

Thank you. What do you think about this particular function? I could document it a little bit better, e.g. with an example of using it with `ghcid` and/or a golden test.

jtdaugherty:
> I don't really know how people typically develop brick apps, perhaps they just run them again every time.

That's what I do, but I don't have any data points on what others do.

> In some cases, especially when working on UI, I think ghcid could provide a very easy and tight feedback loop if one only renders the widget (or whole view) they are working on. If there's other ways to achieve this kind of fast feedback loop, I'd be happy if you'd suggest something.

I guess I am still missing something. How can `ghcid` play any role in using `renderWidget` at all, other than just compiling code that uses it? As far as I know, `ghcid` does one thing and one thing only: it watches for filesystem changes and recompiles an application or library, spitting out compiler warnings and errors, i.e., it does not run the application. Has something significant changed about `ghcid` that I am not aware of?

> This changes enables writing golden tests (for any app using brick), and thus can also allow writing the golden tests for the brick package itself. One benefit of including a few such tests in the package is that they can be used as an example of writing golden tests for other apps.

Okay. I can see how `renderWidget` represents the foundation of a golden test approach, but I am reluctant to merge the patch with that intention unless there's more of a story for golden testing. This function could provide a piece of a testing infrastructure, but I'd want to hear more from you and think more about how test cases could reasonably be captured and how test failures could be reported. I think a challenge with any kind of TUI testing is having a testing API that lets you be expressive enough in your test cases while also being able to update your test case expected outputs easily and present test failures to the user in a way that is digestible.

> Thank you. What do you think about this particular function? I could document it a little bit better, e.g. with an example of using it with ghcid and/or a golden test.

To me, the documentation does not need to be settled quite yet. I wouldn't want the documentation to specifically mention `ghcid` without saying a lot more about how the intended use with `ghcid` can be made easier. The current documentation says that it *can* be used without saying *how*. Ultimately, I'd like the documentation of the function to focus less on broader use cases (which I find are best captured elsewhere like the User Guide) and more on providing a full specification of the function's behavior.

The other thing that came to mind as I looked at the function is that ultimately a testing approach for Brick widgets will need to be able to provide inputs to many of the `RenderState` fields that the function currently defaults to empty. For example, prior to rendering, one might want to queue up a visibility request and then check that the resulting rendering honored it. With that said, I'd want to think about a testing API that allowed those render state fields to default to their current values but provided combinators to build up a declarative spec for a test case that needed to configure specific fields.

Another thing that came to mind is that the return value is already a three-valued tuple, and that could easily grow to four or more fields later as needs evolve. So I'd like to start out with returning a new record type from the function that bundles those fields using meaningful names. That way, new record fields can be added with more ease and less API breakage later.

valyagolev:
> I guess I am still missing something. How can ghcid play any role in using renderWidget at all, other than just compiling code that uses it? As far as I know, ghcid does one thing and one thing only: it watches for filesystem changes and recompiles an application or library, spitting out compiler warnings and errors, i.e., it does not run the application. Has something significant changed about ghcid that I am not aware of?

Yes, `ghcid` has a `--test` argument that allows you to run any function after the compilation and immediately see what it does.
 
    myTestFunc = renderDisplay [ui]

    % ghcid --test Lib.myTestFunc

>  I think a challenge with any kind of TUI testing is having a testing API that lets you be expressive enough in your test cases while also being able to update your test case expected outputs easily and present test failures to the user in a way that is digestible.

I agree that a very beneficial kind of API would be one that allows one to express certain interesting and complex predicates about the rendered result. However, this is far from the goal here, and I don't even think it's possible or convenient to do with what we have here.

The kind of testing I propose here, golden, is not as much about expressing abstract-ish predicates, but about the stability of rendered results. Basically it's like comparing to a screenshot. Typically if one's designs change, they'd rather remove the "screenshot" for that particular view and generate it again. But in other cases – e.g., when working on abstract and stable API functions for UI – you might want to make sure they keep rendering the same, so you keep those screenshots around. It might be useful for making sure some complex layout algorithms don't get broken or whatever.

A testing process would be something along the lines of:

1. One writes a particular `Widget`-returning function that one needs.
2. One makes sure it outputs what one wants (through using it normally in a `brick` app, or even with the `renderDisplay` shortcut).
3. One records a mock-rendered result to a file. This file is unreadable but its stable as long as the function does all the same things.
4. When one evolves the app, the golden part of the test suite makes sure that the output doesn't change when it shouldn't. 
5. If the golden mock output suddenly doesn't match in one of the cases, one might either investigate what happened, or consider that the change was intentional, delete the recorded golden mock output and generate it anew.

> To me, the documentation does not need to be settled quite yet. I wouldn't want the documentation to specifically mention ghcid without saying a lot more about how the intended use with ghcid can be made easier. The current documentation says that it can be used without saying how. Ultimately, I'd like the documentation of the function to focus less on broader use cases (which I find are best captured elsewhere like the User Guide) and more on providing a full specification of the function's behavior.

> The other thing that came to mind as I looked at the function is that ultimately a testing approach for Brick widgets will need to be able to provide inputs to many of the RenderState fields that the function currently defaults to empty. For example, prior to rendering, one might want to queue up a visibility request and then check that the resulting rendering honored it. With that said, I'd want to think about a testing API that allowed those render state fields to default to their current values but provided combinators to build up a declarative spec for a test case that needed to configure specific fields.

> Another thing that came to mind is that the return value is already a three-valued tuple, and that could easily grow to four or more fields later as needs evolve. So I'd like to start out with returning a new record type from the function that bundles those fields using meaningful names. That way, new record fields can be added with more ease and less API breakage later.

What you say makes sense to me as an API for a good testing suite, especially for the internals of the `brick` itself, or a comparably complex framework on top of it. However, I'm not sure if it'd be needed for the apps? I think there's a separate possible direction here, one of simplification, e.g. make this function to be a completely bare-bones:

    renderToPicture :: (Ord n) => V.DisplayRegion -> [Widget n] -> V.Picture

This function has at least three legitimate applications:

1. `ghcid` or otherwisely-driven quick development of the Widgets
2. golden tests
3. to render pretty outputs even without the interactive machinery, e.g. for some kind of reporting. E.g. an app like `hledger` might want to print the output of one of its typically interactive widgets more-or-less verbatim in a batch mode.

I'm not sure if predicates for testing (e.g.) the honouring of visibility requests would be typically useful. I think that as long as `brick` has tested this kind of things, typically the users of the library can probably rely on this functionality as a stable and tested.

I'm only proposing this as an alternative, as a particular different direction that might be beneficial here for the users of `brick`.

valyagolev:
Sorry, closed this as a mistake :)

jtdaugherty:
> Yes, ghcid has a --test argument that allows you to run any function after the compilation and immediately see what it does.

I see, thanks! In that case, something I'd wonder about (even before key event forwarding, as you mentioned) is terminal initialization and control. I imagine `ghcid` assumes that the function prints to `stdout` but does not otherwise change the terminal settings. To view the result of `renderWidget`, you'd need to take over the terminal, show the resulting rendering, and then pause before resetting the terminal state to its prior state, thus effectively erasing the rendering and displaying what was there beforehand. I don't know how that could be done with `ghcid`. (But to me, having a story for that matters before merging this patch if using this functionality with `ghcid` is a stated goal.)

> The kind of testing I propose here, golden, is not as much about expressing abstract-ish predicates, but about the stability of rendered results. Basically it's like comparing to a screenshot.

I'm not proposing that a testing scheme would have abstract predicates; I'm just saying that if a test fails, the failure needs to be understandable. A test without an understandable failure is a frustrating test, and in my experience those tests just tend to get blindly removed and replaced if the failures cannot be understood. I know that dealing with these issues isn't in scope for your patch, but to me, merging this patch without having thought about those issues and planning for them means the patch won't support the intended use case very well. That's why I'm wanting to raise these issues now. Even if the function that gets exposed from `brick` only supports a narrow range of testing use cases, I'd like the documentation to make that really clear to manage expectations.

> What you say makes sense to me as an API for a good testing suite, especially for the internals of the brick itself, or a comparably complex framework on top of it. However, I'm not sure if it'd be needed for the apps? I think there's a separate possible direction here, one of simplification, e.g. make this function to be a completely bare-bones:

Yeah, I think your simplified type signature could be a wrapper around something more sophisticated, with both exported.

> This function has at least three legitimate applications:
>
> ghcid or otherwisely-driven quick development of the Widgets
> golden tests
> to render pretty outputs even without the interactive machinery, e.g. for some kind of reporting. E.g. an app like hledger might want to print the output of one of its typically interactive widgets more-or-less verbatim in a batch mode.

I can believe that these are legitimate applications, but as far as I can tell, they are also all hypothetical at this point, at least based on the issues that it seems need to be considered before these approaches will be reality. Do you have an application that you are developing that needs one of these approaches right now?

> I'm not sure if predicates for testing (e.g.) the honouring of visibility requests would be typically useful. I think that as long as brick has tested this kind of things, typically the users of the library can probably rely on this functionality as a stable and tested.

I don't think that this is the case. While it's true that the low-level implementation of, say, visibility requests in the library need to be correct, there are still plenty of ways in which the behavior of a specific UI is a function of application behavior. For example, an application could simply fail to include `visible` in a particular rendering based on application state. That might be a serious bug. The application's interaction with the renderer occurs in `EventM`, and the record of that interaction is kept in the initial `RenderState` that is fed to the renderer. Side effects induced in `EventM` need to be observable in the rendered result which is why I think (ultimately) a mature testing API for brick needs to have a way that those side effects can be simulated.

valyagolev:
>  I don't know how that could be done with ghcid. (But to me, having a story for that matters before merging this patch if using this functionality with ghcid is a stated goal.)

I tried the setup with `ghcid` before sending the PR, and it functions more or less the way I'd like. (I say "more or less", because 1. there're some artefacts if `DisplayRegion` is too little and 2. I needed to add a `putStrLn ""` in the end to make it work – probably because `vty` resets the cursor to the beginning and `ghcid` somehow mis-scrolls... What I mean is: the setup gets the job done, but it's not extremely pretty and has to be used with caution and understanding of the approach).

>  A test without an understandable failure is a frustrating test, and in my experience those tests just tend to get blindly removed and replaced if the failures cannot be understood. I know that dealing with these issues isn't in scope for your patch, but to me, merging this patch without having thought about those issues and planning for them means the patch won't support the intended use case very well. That's why I'm wanting to raise these issues now. Even if the function that gets exposed from brick only supports a narrow range of testing use cases, I'd like the documentation to make that really clear to manage expectations.

I absolutely agree with you about the frustrating nature of this kinds of tests. As a frequent UI developer, I had to deal with those kinds of frustrating tests way too often. In cases when it's possible and easy, I'd always advocate for doing something different.

That said, those tests get one very particular important job done: they make sure there's no accidental changes in how things look. In my experience, typically those changes amount to just one symbol or a few pixels, changed because of an innocuous-looking rewrite of some general ui-related function... and often there's truly no other way to test for that! So yeah, I find them quite valuable as long as they are easy to use, and everyone involved understands that indeed sometimes nuking a folder of golden samples and generating them anew is just the right thing to do.

Golden UI-related tests do involve a presentation problem before being truly convenient. I have some ideas on how to address that in some cases but ultimately finding the best way to use such tests would be a burden for whoever decides to do that. Writing down the limitations for testing purposes of what we'll end up merging would definitely be valuable. In that case, I'd mention something along the lines of: "You might use the result of the rendering as a base for comparison in your automated test suite. In this case, please bear in mind that vty's mockRenderer will produce a stable, but not completely readable result. You might also try and compare V.Picture's directly." (Actually the last sentence might be the best idea here – I just came up with it – as one can always render both the golden and the actual V.Pictures in the terminal and thus show how they are different).

> I can believe that these are legitimate applications, but as far as I can tell, they are also all hypothetical at this point, at least based on the issues that it seems need to be considered before these approaches will be reality. Do you have an application that you are developing that needs one of these approaches right now?

As I said, I'm using `ghcid` right now with brick for my little app. I haven't pushed it very far just yet. As for the other two use-cases, mostly hypothetical (and that's why I want to keep them simple-but-possible, before we get more feedback).

> I don't think that this is the case. While it's true that the low-level implementation of, say, visibility requests in the library need to be correct, there are still plenty of ways in which the behavior of a specific UI is a function of application behavior. For example, an application could simply fail to include visible in a particular rendering based on application state. That might be a serious bug. The application's interaction with the renderer occurs in EventM, and the record of that interaction is kept in the initial RenderState that is fed to the renderer. Side effects induced in EventM need to be observable in the rendered result which is why I think (ultimately) a mature testing API for brick needs to have a way that those side effects can be simulated.

OK, I see.

Let me say this: for my own purposes (ghcid and maybe simple golden-testing), the simplest `renderToPicture :: (Ord n) => V.DisplayRegion -> [Widget n] -> V.Picture` seems quite enough so far. That's what I'd personally love to see merged. Additionally, I'm happy to work with you to make something more generally useful, as you see fit. What I could probably do is:

1. Provide a prototype/example/a bit of docs of a golden test for a `Widget n`. Since I've already done a bit of trial&error and research on that, I think it can be useful to other people, so if they get a similar idea, they don't need to repeat it to see what they can reasonably achieve.
2. Maybe introduce a few additions to make sure a more general API is useable for testing. As I said, I don't need much of this, so I'll have to use your suggestions to guide myself. From the top of my head: You mentioned introducing a record (`RenderResults`?) to replace the tuple for `(V.Picture, Maybe (CursorLocation n), [Extent n])` (this could also entail changing the type of `renderFinal` to `... -> State (RenderState n) (RenderResults n)`. Exposing either `RenderState(..)` and/or `initialRenderState` would also be quite essential for using `renderFinal`. And perhaps we'd need to derive `Eq` for `RenderState` and `RenderResults` if we want to make them comparable as well, though the users of the library might as well just do this themselves...

jtdaugherty:
> I tried the setup with ghcid before sending the PR, and it functions more or less the way I'd like. (I say "more or less", because 1. there're some artefacts if DisplayRegion is too little and 2. I needed to add a putStrLn "" in the end to make it work – probably because vty resets the cursor to the beginning and ghcid somehow mis-scrolls... What I mean is: the setup gets the job done, but it's not extremely pretty and has to be used with caution and understanding of the approach).

Okay, that helps me to know. The cursor issue makes sense; rendering a `Picture` means placing the cursor as specified in the `Picture`, which might mean not showing one at all or placing it at a particular location.

> "You might use the result of the rendering as a base for comparison in your automated test suite. In this case, please bear in mind that vty's mockRenderer will produce a stable, but not completely readable result. You might also try and compare V.Picture's directly." (Actually the last sentence might be the best idea here – I just came up with it – as one can always render both the golden and the actual V.Pictures in the terminal and thus show how they are different).

That sounds great to me. The `Picture` is ultimately (by the sound of it) what would need to be compared anyway, and it would be pretty easy to write a test driver that lets you interactively compare your golden expected and actual test results!

> Let me say this: for my own purposes (ghcid and maybe simple golden-testing), the simplest renderToPicture :: (Ord n) => V.DisplayRegion -> [Widget n] -> V.Picture seems quite enough so far. That's what I'd personally love to see merged.

I am okay with adding that to the library given how useful it would be to you, and I now believe that there's a story for how the API could be enriched should someone else want a more sophisticated interface for testing. Once you are finished with the patch to your satisfaction, let me know. I'll merge it and adjust style/docs and see about putting out a release.

I really appreciate both the thought you put into this and also the discussion we've had here about some possible issues!

valyagolev:
> I really appreciate both the thought you put into this and also the discussion we've had here about some possible issues!

Thank you, and likewise I appreciate `brick` a lot and am excited to contribute a little.

So I pushed the changes – the simplified function, a bit of docs, and `tests/Render.hs` as an example. It's not ideal what you'd see there, but I think it's a reasonable enough starting point to see if there's any interest in this kind of functionality before introducing more complex machinery. Please let me know if I should add or change anything else.

I'll take a look at the test errors if you approve the workflow for me.

jtdaugherty:
This change is now released in 0.71.

PR #376, opened 2022-07-09
Add Brick.Widgets.Edit.getCursorPosition

TristanCacqueray:


jtdaugherty:
Thanks for this patch! But I just put out Brick 0.72 which adds mouse event handling to editors, which I think means we won't need this. I'd like to hold off on merging it until the need arises again since I am hoping nobody will need this function.

TristanCacqueray:
That's great, thank you! I meant to use this function to implement cut and paste. Would you mind a patch to add that to the `handleEditorEvent`? For example, we could:
- <kbd>ctrl</kbd><kbd>k</kbd> : copy the content into a clipboard before killToEOL
- <kbd>ctrl</kbd><kbd>y</kbd> : paste the content of the clipboard
- Add a new `editorClipboard :: t` field to the Editor struct to keep a copy of the content.

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Bug fixes:
  * Fixed an issue where `tests/Render.hs` did not gracefully exit in the
    presence of an unknown terminal.
 
+API changes:
+ * Added `Brick.Widgets.Edit.getCursorPosition` helper function.
+
```
jtdaugherty, line 14: Just an FYI: please don't add changelog entries. I only add those at release time so I can tell what the appropriate version change should be, and I use the git commit history to help me write the changelog entries for each new release.

jtdaugherty:
This is now released on 0.73.

Issue #379, opened 2022-07-18
Request for comment: EventM state monad refactor


jtdaugherty:
This ticket is a place for discussing the impact of work that is happening on the `refactor/event-state-monad` branch. The objective of this branch is essentially to refactor `EventM` to make event handler code more modular, easier to write, and less verbose. This branch introduces several breaking API changes.

If you write/maintain brick applications, I'd love your thoughts on how this will impact your applications (positively and negatively) if this were to be merged and released. I'd also love feedback if you actually try using the branch as well. Thanks!

Examples of the impact of these changes in action can be found in the demo programs. Here are a few examples:
* MouseDemo: [before](https://github.com/jtdaugherty/brick/blob/master/programs/MouseDemo.hs), [after](https://github.com/jtdaugherty/brick/blob/refactor/event-state-monad/programs/MouseDemo.hs)
* ListDemo: [before](https://github.com/jtdaugherty/brick/blob/master/programs/ListDemo.hs), [after](https://github.com/jtdaugherty/brick/blob/refactor/event-state-monad/programs/ListDemo.hs)
* EditDemo: [before](https://github.com/jtdaugherty/brick/blob/master/programs/EditDemo.hs), [after](https://github.com/jtdaugherty/brick/blob/refactor/event-state-monad/programs/EditDemo.hs)

Here's a summary of the changes and their impacts:

* The `EventM` type is changing from `EventM n a` to `EventM n s a`. `EventM n s a` is now a `MonadState` over `s`. (The state monad introduced in this change is *strict* in `s`.) This means that you can either use lenses or the `mtl` state monad API to manage your state. Therefore, event handler functions that previously took an explicit state argument and returned a `Next` are now state-monadic over that state type instead. For example, a function that was previously written as `handleMyEvent :: s -> BrickEvent n e -> EventM n (Next s)` will now be written `handleMyEvent :: BrickEvent n e -> EventM n s ()`.
* The `Next` type is going away from the user-facing API. This means `EventM` blocks in event handlers no longer need to end with `continue`, `halt`, etc. Instead, the behavior previously provided by the `continue` function is now the default behavior for any `EventM` block, and `halt` etc. need to be called explicitly to indicate what the library's main event loop should do once the event handler terminates. Functions like `halt` do *not* change control flow of `EventM` (i.e. do not cause early aborting).
[...]

PR #380, opened 2022-07-24
WIP: Zoom EventM

xsebek:
This does not (and maybe can not?) work yet! :construction: 

- add `Zoom` for `EventM`:
 ```Haskell
  type instance Zoomed (EventM n s) = Focusing IO
  instance Zoom (EventM n s) (EventM n t) s t
  ```
- add `appState` lens to allow type-changing updates of `EventState` (I have not figured out how to use it though)

- possibly part of #379

### DIFF
--- a/src/Brick/Types/EventM.hs
+++ b/src/Brick/Types/EventM.hs
@@ -1,25 +1,78 @@
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
-module Brick.Types.EventM
-  ( EventM(..)
-  )
-where
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE InstanceSigs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ExplicitForAll #-}
 
-import Control.Monad.Catch (MonadThrow, MonadCatch, MonadMask)
+module Brick.Types.EventM (
+  EventM (..),
+) where
+
+import Control.Monad.Catch (MonadCatch, MonadMask, MonadThrow)
 import Control.Monad.Reader
 import Control.Monad.State.Strict
 
 import Brick.Types.Internal
 
+import Control.Monad.Identity (Identity)
+import Lens.Micro
+import Lens.Micro.Internal
+import Lens.Micro.Mtl.Internal
+import Lens.Micro.Mtl
+
 -- | The monad in which event handlers run.
-newtype EventM n s a =
-    EventM { runEventM :: ReaderT (EventRO n) (StateT (EventState n s) IO) a
-           }
-           deriving ( Functor, Applicative, Monad, MonadIO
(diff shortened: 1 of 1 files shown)


xsebek:
The microlens documentation for [zoom (see note at the bottom)](https://hackage.haskell.org/package/microlens-mtl-0.2.0.2/docs/Lens-Micro-Mtl-Internal.html#v:zoom) suggests that this should be really easy for newtypes.

But for me, the lens types are hard to read.

---

To make this more concrete here is the `zoom` signature as I think it would be used in swarm:
```Haskell
zoom :: Lens' AppState UIState -> EventM Name UIState () -> EventM Name AppState ()
```
but in reality `zoom` is more general:
```Haskell
zoom :: LensLike' (Focusing IO c) s s -> EventM n s c -> EventM n s c
```
which makes it hard for me to understand how that lens-like thing should be applied to the state in `EventState`. :confused: 

I hoped it would be as easy as composing the `appState` with the lens-like thing, but it gives unreadable type errors:
```Haskell
unfocusing #. (appState . l) (Focusing #. m)
-- Occurs check: cannot construct the infinite type:
--  s ~ EventState n s
```

Any ideas @jtdaugherty?

xsebek:
I tried putting explicit types and seeing where it goes wrong:
```Haskell
instance Zoom (EventM n s) (EventM n t) s t where
  zoom :: LensLike' (Focusing IO c) t s -> EventM n s c -> EventM n t c
  zoom l (EventM evm) = EventM (zoomOnReader evm)
   where
    zoomOnReader :: ReaderT e (StateT (EventState n s) IO) c -> ReaderT e (StateT (EventState n t) IO) c
    zoomOnReader (ReaderT rm) = ReaderT (zoomOnState . rm)
    zoomOnState :: StateT (EventState n s) IO c -> StateT (EventState n t) IO c
    zoomOnState (StateT sm) = StateT $ uf l sm 
    uf
      -- we get a type-changing lens thing
      :: LensLike' (Focusing IO res) old new
      -- then we get the original action with old type which we are changing to the new one
      -> (EventState n old -> IO (res, EventState n old))
      -- finally, we get a nice new action that we can pass to StateT
      -> EventState n new -> IO (res, EventState n new)
    uf l' m =
      -- we get the result out of Focusing
      unfocusing .
        -- we focus on the inner state and change type     -- DOES NOT TYPECHECK
        ((appState . l') :: LensLike' (Focusing IO res) (EventState n old) (EventState n new))
          -- we put the result into Focusing               -- DOES NOT TYPECHECK
          ((Focusing . m) :: EventState n old -> Focusing IO res (EventState n old))
```
My intuition is that _**this should work**_, but I can not understand the expanded lens signatures to see where the problem is.

Maybe I could try again with the [`optics` library](https://hackage.haskell.org/package/optics-extra-0.4.2.1/docs/Optics-Zoom.html) that has more OOP style types and could have easier-to-understand error messages? :thinking: But I have not used it in a while. :confused:

xsebek:
Alternatively, maybe the easiest path would be to put the `s` state into a separate `StateT` transformer from the rest of `EventState`.

That could possibly allow the instance to be as simple as:
```Haskell
zoom l (EventM m) = EventM (zoom m) -- basically what microlens was suggesting
```

But that would make it a bit harder to use in the rest of the library. Possibly the damage could be mitigated by writing the getter of `EventState` from `EventM` so that it puts together the `s` state and the rest from two `StateT` transformers.

jtdaugherty:
> But that would make it a bit harder to use in the rest of the library. Possibly the damage could be mitigated by writing the getter of EventState from EventM so that it puts together the s state and the rest from two StateT transformers.

I'm thinking this shouldn't matter; the library internals already need to go around the `MonadState` instance to manipulate `EventState` anyway. The `MonadState` instance is part of the public API and should only ever be in terms of `s`, so I think it shouldn't matter where in the transformer stack `s` sits. The internals will (and do) need to work around that either way.

Thank you for working on this! I'll check back later (I'm away most of the rest of the day).

jtdaugherty:
I got this working in dee90efa8cde1755f435c1f7dd9b6828b1bd0089 so this PR can be closed. Thank you for your work on this!

Issue #356, opened 2021-12-26
Wide characters taking up more space than needed on border labels


iogrt:
Using emojis inside border labels is making the box take up 1 more character than needed.

Example:
![2021-12-26_19:41:52](https://user-images.githubusercontent.com/39342422/147419569-5f776035-7de5-4d1c-b28a-6bbd9c8274de.png)

```hs
simpleMain (borderWithLabel (str "🌀 test string") (vBox $ str <$> inits "HelloWorld"))
```

jtdaugherty:
Thank you for the report! The state of wide unicode characters in terminals is not great. Brick (and its driver library, Vty), only have a little bit of control over how this works, unfortunately. You can read some details about this [here](https://github.com/jtdaugherty/vty/issues/175). Brick does its best to calculate the column width of Unicode characters, but it gets many of them wrong (because Vty gets them wrong). And Vty can't always get them right because even the terminal emulator must agree with Vty, and many terminal emulators do not even get the widths right.

So for now, I encourage you to stick to characters that don't suffer from this problem. (Everywhere in Brick UIs, not just border labels.)

iogrt:
Thanks for the quick response, and sorry for the duplicate issue. I've been reading the threads that you have mentioned and understand why it's such a hard thing to get right.

Since I'm developing these scripts for myself, and for anyone who comes across this, I've found a way to tell Brick what size I have for my emojis (which is 2 width instead of one, for `ghci`):
```hs
-- imports
import qualified Graphics.Vty.Image.Internal as VI (Image (HorizText))
-- code
emoji s = raw $ VI.HorizText V.defAttr (L.pack s) 2 1
myApp = simpleMain (borderWithLabel (emoji "🌀" <+> str "test string") (vBox $ str <$> inits "HelloWorld"))

```

jtdaugherty:
Thanks! That seems like it could work well for specific characters. Be sure to test on different terminal emulators, though; the width that you hard-coded must agree with what the emulator thinks, and that may vary, especially for characters more recently added to the Unicode standard.

jtdaugherty:
(Also, since you used an "internal" Vty API, that is not guaranteed to be stable.)

Issue #265, opened 2020-03-02
Border broke due to emojis/wide characters in list mode


hyperreal-gopher:
I read [the section on wide characters](https://github.com/jtdaugherty/brick/blob/master/docs/guide.rst#wide-character-support-and-the-textwidth-class), but I think your borders might be broken if the line includes a wide character.

Here's a screenshot: https://cf.mastohost.com/v1/AUTH_91eb37814936490c95da7b85993cc2ff/fosstodon/media_attachments/files/004/026/057/original/f85ef423cb31c9f9.png

Thank you for your time!

Garmelon:
This seems to happen with some emojis, but not all wide characters. Here is a very small test script:
```haskell
import Brick
import Brick.Widgets.Border

widget :: Widget ()
widget = vBox
  [ border $ str "📁"
  , border $ str "⼔"
  ]

main :: IO ()
main = simpleMain widget
```
![image](https://user-images.githubusercontent.com/11077553/75669800-95a78680-5c73-11ea-8010-97f8c922f1f1.png)

Garmelon:
Since the guide says that brick relies on Vty's use of `utf8proc`, this is what utf8proc seems to think is the correct width:
```c
#include <stdio.h>
#include <utf8proc.h>

const utf8proc_int32_t FOLDER = 0x1f4c1;
const utf8proc_int32_t WIDECHAR = 0x2f14;

int main(void) {
  printf("Width of \xf0\x9f\x93\x81 is %i.\n", utf8proc_charwidth(FOLDER));
  printf("Width of \xe2\xbc\x94 is %i.\n", utf8proc_charwidth(WIDECHAR));
  printf("Width of A is %i.\n", utf8proc_charwidth('A'));
}
```
![image](https://user-images.githubusercontent.com/11077553/75671826-5b3fe880-5c77-11ea-86be-5d243d7c4cb9.png)

Garmelon:
And a final test: Unsurprisingly, Vty reports the character widths incorrectly:

```haskell
import Graphics.Vty

printWidth :: Char -> IO ()
printWidth c = putStrLn $ "Width of " ++ [c] ++ " is " ++ show (wcwidth c) ++ "."

main :: IO ()
main = mapM_ printWidth ['📁', '⼔', 'A']
```
![image](https://user-images.githubusercontent.com/11077553/75673383-bfb07700-5c7a-11ea-83ef-a8eed6c8c3db.png)

Tested with vty 5.26.

[...]

PR #381, opened 2022-07-26
Update documentation for wide character support

avh4:
The guide says "Brick relies on Vty's use of the [utf8proc](http://julialang.org/utf8proc/) library to determine the column width of each character rendered." but use of utf8proc was removed in vty-5.13  https://github.com/jtdaugherty/vty/commit/cd534c1028b3efde9c13218a10b7e085958496bc.

This updates the brick guide to point to how to configure the new way of computing character widths that was since added.  (It took me a few hours of digging through issue histories to find this and understand that it was the current correct way!)

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -7,3 +7,23 @@ brick FAQ
   default and requires configuration to make it work. See also:
 
   http://unix.stackexchange.com/questions/110022/how-do-i-find-out-the-keycodes-for-ctrlup-and-down-arrow-for-term-screen
+
+* Q: Why do some emojis mess up the layout?
+* A: For wide characters to be displayed correctly, [vty]'s
+  determination of the character width and the user's
+  terminal emulator's determination of the character width
+  must match. Unforunately, every terminal emulator
+  calulcates this differently, and none correctly follow
+  the Unicode standard.
+  The issue is further complicated by Unicode combining
+  characters and releases of new versions of the Unicode
+  standard. 
+
+  As a result, the current recommendation is to avoid
+  use of wide characters due to these issues.
+  If you still must use them, you can read [vty]'s
+  documentation for options that will affect character
+  width calculations.
+
+
+[vty]: https://hackage.haskell.org/package/vty
--- a/docs/guide.rst
+++ b/docs/guide.rst
@@ -921,14 +921,16 @@ with ``Brick.Themes.saveCustomizations``.
 Wide Character Support and the TextWidth class
 ==============================================
 
-Brick supports rendering wide characters in all widgets, and the brick
-editor supports entering and editing wide characters. Wide characters
-are those such as many Asian characters and emoji that need more than
-a single terminal column to be displayed. Brick relies on
-`vty's multi-column character support`_ to determine the column width of
-each character rendered, but this requires some additional
-configuration.  Read the details there to configure the correct widths
-for your terminal.
+Brick attempts to support rendering wide characters in all widgets,
+and the brick editor supports entering and editing wide characters.
+Wide characters are those such as many Asian characters and emoji
+that need more than a single terminal column to be displayed.
+
+Unfortunatley, there is not a fully correct solution to determining
+the character width that the user's terminal will use for a given
+character. The current recommendation is to avoid use of wide characters
+due to these issues. If you still must use them, you can read `vty`_'s
+documentation for options that will affect character width calculations.
 
 As a result of supporting wide characters, it is important to know that
 computing the length of a string to determine its screen width will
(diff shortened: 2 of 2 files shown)


avh4:
For reference, past related issues:
- https://github.com/jtdaugherty/brick/issues/265
- https://github.com/jtdaugherty/brick/issues/356
- https://github.com/jtdaugherty/vty/issues/175 (implements the current solution)

xsebek:
Interesting, should applications using brick and wide characters suggest users create the configuration for their terminal?

Or would it be best to give an option like `--unicode iterm2` and have the brick application generate it for the user? :confused: 

The Matterhorn application mentioned in one of those issues does not suggest a solution in its FAQ:
- https://github.com/matterhorn-chat/matterhorn/blob/master/docs/FAQ.md#does-matterhorn-support-graphical-emoji

jtdaugherty:
Unfortunately it's a little worse than the current state. When I added the "width table" feature to Vty, it seemed at the time that it could result in width tables that were on average more likely to be correct than the width table that was built into the library at the time. That turned out to be quite wrong; the process of querying the terminal got wrong answers sometimes, mostly because the character ranges that get queried include combining characters. So the width table approach itself is actually not going to work in the long run because a proper width calculation approach needs to perform Unicode-aware lookahead. This is because there are lots of Unicode character sequences that need to be combined *before* their width can be determined.

So at this point it's not quite right to say that the table generation approach is the best way to go; we provided it in case someone found that, in their situation, it gave a better answer than the built-in (old) table.

With that said, I'm torn on what to do about it because I can't recommend that people always use the table generator, but clearly the old built-in table will be wrong for wide characters (or combinable ones) added in new revisions of Unicode. We also don't know in general how smart the terminal's Unicode width calculations are, and we've found that in some cases they have just as naive an approach as Vty has without the tables set up (i.e. a fixed table with no lookahead).

My recommendation at this point is to avoid wide characters because in my view, wide character support in terminals in general is just a huge mess. We've tried hacks in Vty to compensate, but they break as much as they fix because different terminals are at different levels of broken, and those brokenness levels change over time as Unicode is updated!

As far as this patch is concerned, I don't want Brick's documentation to say anything about the Vty generated table configuration quite yet because (as I said above) I wouldn't recommend that be the first approach people take. I'd be okay with the Brick docs just saying that Brick depends on Vty's *API* for wide characters. Under the hood that API could use different mechanisms independently.

Thanks!

avh4:
So the situation I'm trying to improve is the one I was in:
- I started using brick
- eventually I tried putting a few emojis in the UI (in my case it was 😵 and 🌘)
- the layout broke from the emoji
- I searched for issues and documentation and eventually read that things were supposed to work
- I confirmed I had the latest brick and vty versions
- things still didn't work, so I had to re-read all the existing issues -- but it's quite confusing because you eventually realized that there were different fixes at different times and you have to start looking at the timestamps of the comments to understand the sequence of events and the current state of things
- I even had to figure out how to write a little c program to determine whether or not utf8proc was working correctly because I thought it was meant to be being used :sweat_smile:
- I also tried the workaround mentioned in https://github.com/jtdaugherty/brick/issues/356#issuecomment-1001247381, which seemed to work at first, but later I found it caused bugs with screen redrawing

Anyway, as soon as I ran `vty-build-table-width -u`, everything worked fine for me (so far, at least...).  And I'm also planning to have my app use the table-building API on first startup before I release it.

So I'd like to propose at least one of the following:
- I could modify this PR to still link there, but to only point you there if you run into issues, and not as a general recommendation (or maybe even add your quote saying you officially don't  recommend it)
- cancel this PR, and I could just add a note as a comment to the past issues I found when looking into this
- slightly modify the default behavior so that emoji default to 2 width instead of 1 width (like 😵 and 🌘)

Would the first of those (link but with a warning) be okay?

jtdaugherty:
@avh4 thanks for such a detailed explanation of how you got to this point. It's both helpful for me to hear it, and saddening to hear that it was way more indirect than I'd like. Since it's also no fun for me to apologize for the situation several times a year (you aren't the first person to encounter these issues), it would also help me to set some expectations about this more clearly in the documentation. It seems like an FAQ entry as well as some User Guide material on this would have been helpful. I'm happy to make some updates along those lines.

As for this patch, your first option works for me. Thank you!

avh4:
Okay, I updated it to hopefully make the current recommendation clear, and also added it to the FAQ 8baa49f29b4f92e3211e754386e2764ef378de32

How does that look?

jtdaugherty:
Looks good to me. Thank you!

PR #382, opened 2022-07-28
add listSelectedElementL Traversal

frasertweedale:
Add a `Traversal` that targets the selected element (if any).  It is
useful for retrieving, setting, modifying or traversing the selected
element and composes with other optics.  Asymptotics are documented.

It also makes it possible to rewrite some other functions in terms
of `listSelectedElementL`, including:

- `listModify`: Improved asymptotics for some containers, including
  `Seq`.  Introduces `Splittable` and `Semigroup` constraint.

- `listSelectedElement`: Simplifies definition.  Asymptotics
  unchanged.  Introduces `Semigroup` constraint.

Such changes are left for discussion of possible subsequent
implementation.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -37,6 +37,7 @@ module Brick.Widgets.List
   , listSelectedL
   , listNameL
   , listItemHeightL
+  , listSelectedElementL
 
   -- * Accessors
   , listElements
@@ -81,7 +82,7 @@ import Control.Applicative ((<|>))
 import Data.Foldable (find, toList)
 import Control.Monad.Trans.State (evalState, get, put)
 
-import Lens.Micro ((^.), (^?), (&), (.~), (%~), _2, _head, set)
+import Lens.Micro (Traversal', (^.), (^?), (&), (.~), (%~), _2, _head, set)
 import Data.Functor (($>))
 import Data.List.NonEmpty (NonEmpty((:|)))
 import Data.Maybe (fromMaybe)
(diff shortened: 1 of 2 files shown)


src/Brick/Widgets/List.hs
```diff
@@ -602,6 +603,31 @@ listFindBy test l =
...
+-- listSelectedElementL for 'List': O(1) -- preview, fold
+--                                O(n) -- set, modify, traverse
+-- listSelectedElementL for 'Seq.Seq': O(log(min(i, n - i)))  -- all operations
+-- @
+--
+listSelectedElementL
+  :: (Splittable t, Traversable t, Semigroup (t e))
+  => Traversal' (GenericList n t e) e
+listSelectedElementL f l =
+  case l ^. listSelectedL of
```
jtdaugherty, line 620: I'll take care of this change in this case, but I wanted to mention for future PRs: it would help me if the style of the patches matches the style of the surrounding code, i.e. four-space indentation and `let` style as follows:
```haskell
let foo = ...
    bar = ...
in ...
```

jtdaugherty:
Thanks for this change and for the detailed commit message! (That will make the release changelog update very easy.)

frasertweedale:
@jtdaugherty no worries.  Sorry I didn't pay much attention to the style; it was a cut-and-paste from another project.

What are your thoughts on refactoring `listModify` and `listSelectedElement` to use `listSelectedElementL` under the hook?

jtdaugherty:
Oh, I had read about that in the commit message and missed the fact that it was described as hypothetical rather than part of the patch. I'd be happy to accept a PR to refactor the existing functions provided they do not change externally-visible behavior in any way. If you're up for writing that, that works for me - or if not, let me know.

frasertweedale:
@jtdaugherty thanks, I'll do it soon.

PR #390, opened 2022-09-19
Add ghc 9.2.4 to github CI config

jhrcek:
....Hello,
I updated tested-with in cabal file to latest available patch version for each minor versions of GHC and regenerated github CI config by running `haskell-ci regenerate`. The main idea is to extend CI by adding the most recent minor versions of ghc 9.

EDIT: seems like 9.4.2 won't be that simple to add, because cabal can't find a build plan (config-ini doesn't allow text 2.0 dependency). So for now at least bumping patch versions of GHCs in CI..

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -32,11 +32,6 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


jtdaugherty:
Thank you!

PR #394, opened 2022-09-22
Fix haddock examples

jhrcek:
Seems like these were forgotten during removal of instance IsString AttrName..


jtdaugherty:
Thanks!

PR #402, opened 2022-10-15
renderBox: avoid redundant cropResultToContext calls

frasertweedale:
In profiling output I noticed that `renderBox` crops the child widgets three times.  For example:

```
                                       individual      inherited
                            entries  %time %alloc   %time %alloc

vLimit                        32637    1.4    1.5    10.4   11.5
 cropResultToContext          32637    0.4    1.5     1.4    2.6
  crop                        32637    0.9    1.0     0.9    1.0
   setCoordinates             32637    0.0    0.0     0.0    0.0
  crop                        32637    0.0    0.1     0.2    0.1
   cropBottom                 32637    0.2    0.0     0.2    0.0
    imageHeight               32637    0.0    0.0     0.0    0.0
   cropRight                  32637    0.0    0.0     0.0    0.0
    imageWidth                32637    0.0    0.0     0.0    0.0
  availHeight                 32626    0.0    0.0     0.0    0.0
 hLimit                           0    0.9    1.3     7.5    7.5
  cropResultToContext         65274    2.0    2.9     5.9    5.2
 …
```

Note the calls to `cropResultToContext`.  It is called once for each call to the enclosing `vLimit`, and a further two times in the inner `hLimit` call.  In the implementation, that pattern that causes this is:

```haskell
render $ limitPrimary br remainingPrimary
       $ limitSecondary br availSecondary
       $ cropToContext widget
```

`limitPrimary` and `limitSecondary` are `vLimit` and `hLimit` or vice-versa, depending on whether we are rendering a `vBox` or `hBox`. Both `vLimit` and `hLimit` modify the available space in the render context and call `cropToContext`, so the explicit `cropToContext` is redundant.  The `limitSecondary` is also redundant, because `availSecondary` is taken directly from the render context.  So, remove the redundant operations.

As a result of this change, total program allocation, and allocation for the `vBox` cost centre that is the parent of the cost centres table shown above, improve from:

```
total time  =        0.61 secs   (614 ticks @ 1000 us, 1 processor)
total alloc = 856,554,568 bytes  (excludes profiling overheads)

                                 individual      inherited
                      entries  %time %alloc   %time %alloc

vBox                        0    5.0    9.0    23.9   25.9
```

to

```
total time  =        0.61 secs   (606 ticks @ 1000 us, 1 processor)
total alloc = 773,138,088 bytes  (excludes profiling overheads)

                                 individual      inherited
                      entries  %time %alloc   %time %alloc

vBox                        0    5.3    8.9    19.0   19.9
```

Exactly the same number of `vLimit` calls occurs; this is the profiling caused by navigating around a specific, large mail message in Purebred. Total program CPU time and allocation have non-deterministic factors, so take them with a grain of salt.

Allocations in `renderBox` are reduced significantly for long lists of widgets.  In my scenario (rendering and navigating around a long email in Purebred), this changes eliminates ~10% of all program allocation.

Surprisingly, although the cost centres report lower CPU time, the total program CPU time did not change significantly.  I don't have a theory to explain this.  But at least it's not a regression.

### DIFF
--- a/src/Brick/Widgets/Core.hs
+++ b/src/Brick/Widgets/Core.hs
@@ -639,18 +639,13 @@ renderBox br ws =
           (his, lows) = partition (\p -> (primaryWidgetSize br $ snd p) == Fixed)
                         pairsIndexed
 
-      let availPrimary = c^.(contextPrimary br)
-          availSecondary = c^.(contextSecondary br)
-
           renderHi prim = do
             remainingPrimary <- get
-            result <- lift $ render $ limitPrimary br remainingPrimary
-                                    $ limitSecondary br availSecondary
-                                    $ cropToContext prim
+            result <- lift $ render $ limitPrimary br remainingPrimary prim
             result <$ (put $! remainingPrimary - (result^.imageL.(to $ imagePrimary br)))
 
       (renderedHis, remainingPrimary) <-
-        runStateT (traverse (traverse renderHi) his) availPrimary
+        runStateT (traverse (traverse renderHi) his) (c ^. contextPrimary br)
 
       renderedLows <- case lows of
           [] -> return []
@@ -660,10 +655,7 @@ renderBox br ws =
                   primaries = replicate rest (primaryPerLow + 1) <>
                               replicate (length ls - rest) primaryPerLow
 
-              let renderLow ((i, prim), pri) =
-                      (i,) <$> (render $ limitPrimary br pri
-                                       $ limitSecondary br availSecondary
-                                       $ cropToContext prim)
+              let renderLow ((i, prim), pri) = (i,) <$> render (limitPrimary br pri prim)
 
               if remainingPrimary > 0 then mapM renderLow (zip ls primaries) else return []
 


frasertweedale:
@jtdaugherty fyi am working towards further performance improvements in various areas of brick.

jtdaugherty:
Thanks for your work on this, as well as the detailed explanation. I'm going to merge it, and while I do trust your findings, I also want to just express skepticism about time and allocation profiles in general. In my experience it's always difficult to know how comparable two profiles are because comparing them depends on taking similar enough runtime code paths. That's difficult to ensure unless your profiles come from a test suite or other automated/deterministic way of driving the program. In your case, were the profiles obtained by just providing more or less the same user input to two program runs?

frasertweedale:
@jtdaugherty a fair question.  In this case, I am displaying the same large email (2967 lines) and providing the same inputs to perform a substring search, and navigate to the next 10 matches in the mail body.

The mail has 2965 lines and gets rendered with two additional lines of frontmatter.  Each line being a separate widget in a `vBox`.  So, that's 11 times executing the "highlight *n*th search result" render routine.  So that's 11 * (2965 + 2) = 32637 times doing whatever we do to each child widget of the `vBox`.   If you look in both profile outputs (before/after attached), you'll see that number.  And in the first case, some multiples of 32637 with substantial costs beside them.

I agree that deterministic results are important.  While there plenty of non-deterministic behaviour in purebred giving rise to variation in total program costs, the paths and cost centres analysed are reached deterministically, so the results are reliable.

- [renderBox-crop-0.prof.txt](https://github.com/jtdaugherty/brick/files/9792626/renderBox-crop-0.prof.txt)
- [renderBox-crop-1.prof.txt](https://github.com/jtdaugherty/brick/files/9792627/renderBox-crop-1.prof.txt)

jtdaugherty:
Okay, cool, thanks!

It also occurs to me that if your current implementation is using boxes heavily for things that have a well-known size (i.e. all 1 line high) then you might consider using a custom widget to concatenate the underlying images together rather than using `vBox` and `hBox`. That'll be faster, but whether it would entail giving up some functionality depends on whether you're embedding e.g. clickable regions in those boxes. If you aren't -- if you're just highlighting words/tokens, etc. -- then you could probably concatenate the results yourself and get a performance boost just by not using boxes at all (or nearly as much).

Of course, with that said, using boxes less would also lead to you finding and fixing performance issues less often.. :)

jtdaugherty:
(And the other thing I feel the need to mention is that if your email contents aren't changing while you view them, then render-caching the result would also be worth considering.)

frasertweedale:
@jtdaugherty thank you for the tips.

We already cache the image, but we re-render when we highlight, and when we "go to next match" (because "current match" is highlighted differently from "other match".

As for concatenating the images ourselves - we'll look into that!

jtdaugherty:
This change is now released in `1.4`.

PR #413, opened 2022-11-24
Simplify renderBottomUp

mlang:
* Use compare instead of nested ifs
* Use vertJoin instead of vertCat with a list of two elements


jtdaugherty:
Thanks for this patch! While I can see how this could be seen as a better implementation in some sense, I'm going to decline it on readability grounds. In the demo programs I'm striving for maximum clarity and sometimes that means writing things out in a less compact but more explicit way.

PR #415, opened 2022-12-02
Added haskell tic-tac-toe

BlastWind:
Thanks Jonathan, as a Haskell beginner, I learned a lot from using `bricks`.  After I get more experience in Haskell, I think I'll even make a code base deep dive video on `bricks` and `vty`, if you don't mind.

Question: Do you have any general tips, or know of good open source projects I can read to help me add LAN multiplayer functionalities to this game? I'm thinking of using the primitive [Network.Socket](https://hackage.haskell.org/package/network-3.1.2.7/docs/Network-Socket.html).


jtdaugherty:
Thanks, I'm glad you found `brick` useful and it'd be really cool if you ever made such a video.

As far as network APIs go, you could go lots of different ways but if it were me, I'd probably use something like [zeromq4-haskell](https://hackage.haskell.org/package/zeromq4-haskell) rather than the low-level socket API. As far as projects to check out that do something similar, I don't know of any off hand but you might see if any of the projects in the brick README do network stuff. Maybe a few do.

PR #418, opened 2022-12-17
Add haradict to applications using brick.

srhoulam:
Got another brick app for you. Thanks again for making brick!


jtdaugherty:
Cool! I hope you're having a good experience using the library and please let me know if you have questions/feedback.

PR #444, opened 2023-02-20
Fix documentation for handleListEventVi

amano-kenji:
I think
```
'handleListEventVi' 'handleListEvent'
```
was a mistake.


jtdaugherty:
Thanks!

PR #449, opened 2023-03-08
Add a function to create an attribute from a specified style

adql:
The new function `Brick.Util.style` is analogous to the existing `fg` and `bg` functions.

Also fix a small wording mistake in `bg`'s Haddock.


jtdaugherty:
Thank you!

PR #458, opened 2023-04-13
[GHC-58520] [-Wtype-equality-requires-operators]

mlang:
The use of ‘~’ without TypeOperators will become an error in a future GHC release


jtdaugherty:
Thanks!

PR #463, opened 2023-05-06
simplify FileBrowser selection code, fix bracket lints

kostmo:


jtdaugherty:
Hi @kostmo, thanks for your patch. Could you share a bit about what motivated these changes? Was this the result of running a linter, or did you want to make these changes as a result of looking at the code to understand something?

kostmo:
> what motivated these changes

Hi, a little of both---for swarm-game/swarm#1010 I want to configure the `FileBrowser` dialog to close upon selection of just a single file, rather than marking multiple files.  So I found myself studying the `maybeSelectCurrentEntry` function, which was a bit cumbersome to read.  Meanwhile the linter was yelling to fix parentheses in the same file :).  Ultimately I found a technique [in the FileBrowser demo](https://github.com/jtdaugherty/brick/blob/a797c86da54a7017a7458b113eb40ac7f15f1de3/programs/FileBrowserDemo.hs#L67-L75).

jtdaugherty:
Okay, thanks for the context!

I'll merge this, but I have a request for future patches, if you submit any: please make atomic commits that make one logical (type of) change per commit. Concretely, I'd have liked the the single commit in this PR to be two commits: one to change only the parens, and one to refactor existing code. Commits that make multiple changes that are unrelated to each other can be more difficult to review *and* are not as amenable to `git bisect` in case of a future problem. If you don't know, `git add -p` is the best way to do this if you find yourself making unrelated changes to the same file; that will prompt per-hunk so you can stage just the right changes and leave others behind for subsequent commit(s).

PR #483, opened 2023-08-20
Updating to support building on Windows.

chhackett:
Here's my fix for the FileBrowser module. It does use my favorite trick for cross-compilation, the #ifdef.

I'm not sure how you'd like to handle cross platform support for this. If you have suggestions for improvement, I'm certainly willing to refactor this...

### DIFF
--- a/brick.cabal
+++ b/brick.cabal
@@ -114,9 +114,10 @@ library
                        text-zipper >= 0.13,
                        template-haskell,
                        deepseq >= 1.3 && < 1.5,
-                       unix,
                        bytestring,
                        word-wrap >= 0.2
+  if !os(windows)
+    build-depends:     unix
 
 executable brick-custom-keybinding-demo
   if !flag(demos)
--- a/brick.cabal
+++ b/brick.cabal
@@ -115,9 +115,8 @@ library
                        template-haskell,
                        deepseq >= 1.3 && < 1.5,
                        bytestring,
-                       word-wrap >= 0.2
-  if !os(windows)
-    build-depends:     unix
+                       word-wrap >= 0.2,
+                       unix-compat
 
 executable brick-custom-keybinding-demo
   if !flag(demos)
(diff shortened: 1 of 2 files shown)


jtdaugherty:
Thank you for this patch! Thinking about it a bit more now that it's it front of me, I really wonder if there's a way to do this in a platform-agnostic way. I really want to make Brick cross-platform by not using any platform-specific stuff in it (going forward) rather than ifdef-guarding platform-specific implementations of things. (Since I don't have the resources/desire to support Windows specifically, I hope to get support for it for free by using platform-agnostic libraries.)

Would you be willing to investigate whether there's a more platform-agnostic way to write this code that doesn't require ifdefs in the Brick implementation? Perhaps there is a library that already handles these details.

chhackett:
@jtdaugherty 
Yup, I'll keep looking into it. But one issue I see right away is the enumeration of FileType(s):

-- | The type of file entries in the browser.
data FileType =
    RegularFile
    -- ^ A regular disk file.
    | BlockDevice
    -- ^ A block device.
    | CharacterDevice
    -- ^ A character device.
    | NamedPipe
    -- ^ A named pipe.
    | Directory
    -- ^ A directory.
    | SymbolicLink
    -- ^ A symbolic link.
    | UnixSocket
    -- ^ A Unix socket.
    deriving (Read, Show, Eq)

Windows has no notion of BlockDevice, CharacterDevice, NamedPipe, or UnixSocket. So I don't think there will be a platform independent way to recover those values on *nix platforms. The directory package basically has functions to determine if a file is a directory, a regular file, or a symlink for example.

I've looked around for other packages in the past and haven't found anything better. So the best we would be able to do in an agnostic way is just to determine if it is either a file, directory, symlink, or 'something else'.

chhackett:
I was wrong. I think I found the package I need: unix-compat.

I'll update the PR hopefully today. But maybe later this week.

chhackett:
Turned out to be a rather simple change...

jtdaugherty:
Wonderful, I will take a look at the latest commits later today!

jtdaugherty:
This looks great to me. Thanks for your work on this!

PR #487, opened 2023-10-31
Make it build with ghc 9.8 (update deps only)

erikd:
I had to add a `cabal.project`  file containing:
```
packages:
  .

allow-newer:
  , data-clist:deepseq
  , config-ini:containers
  , config-ini:text
```
I already have PRs open for those two packages.

This can be fixed with a Hackage metadata edit.


jtdaugherty:
Thanks! Metadata for 2.0 is now updated also.

PR #508, opened 2024-05-07
Remove trailing slash from FileBrowser initial directory specification.

kquick:
If the FileBrowser is initialized with a FilePath that ends in a slash, then if the user hits Enter on the "../" entry to move to the parent directory, the only effect is the removal of that trailing slash.  This is relatively unexpected and likely to be unnoticed by the user, possibly leading them to believe that the move-to-parent-directory operation did not work.

Once the trailing slash is removed, the expected move occurs whenever the "../" entry is invoked, and the current path never ends in a trailing slash.

This patch simply removes the trailing slash(es) from the user-supplied FilePath (although it will leave "/" untouched).

### DIFF
--- a/src/Brick/Widgets/FileBrowser.hs
+++ b/src/Brick/Widgets/FileBrowser.hs
@@ -154,7 +154,7 @@ import qualified Data.Text as T
 import Data.Monoid
 #endif
 import Data.Int (Int64)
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Thank you!

jtdaugherty:
This is now released in 2.3.2.

PR #511, opened 2024-06-29
Replace keybinding name fromJust with maybe

xsebek:
When I forgot a `KeyEvents` entry and tried to print all keybindings, the `fromJust` failed with an error that does not show the problematic handler.

Instead of erroring out, write `(unnamed)` into Event Name, so that it's easy to see which handler is missing a name.

### DIFF
--- a/src/Brick/Keybindings/Pretty.hs
+++ b/src/Brick/Keybindings/Pretty.hs
@@ -25,7 +25,6 @@ where
 
 import Brick
 import Data.List (sort, intersperse)
-import Data.Maybe (fromJust)
 #if !(MIN_VERSION_base(4,11,0))
 import Data.Monoid ((<>))
 #endif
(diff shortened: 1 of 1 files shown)


src/Brick/Keybindings/Pretty.hs
```diff
@@ -124,7 +124,7 @@ mkKeybindEventHelp kc h =
           ByKey b ->
               (Comment "(non-customizable key)", [Verbatim $ ppBinding b])
           ByEvent ev ->
-              let name = fromJust $ keyEventName (keyConfigEvents kc) ev
+              let name = fromMaybe "<unnamed>" $ keyEventName (keyConfigEvents kc) ev
```
xsebek, line 127: @jtdaugherty I don't understand why this failed for me - I do not see `<unnamed>` in the [output](https://github.com/swarm-game/swarm/pull/1979#issuecomment-2195592963).

xsebek: I must have missed one event in `KeyEvents` and then fixed it before creating this PR.

I deleted one, and the text is shown now:
```markdown
| Keybinding | Event Name | Description |
| ---------- | ---------- | ----------- |
| `C-q` | (unnamed) | Open quit game dialog |
```

jtdaugherty:
I'm a little confused - could you provide some context for this patch? The commit message does not provide a motivation for it, so I don't know what problem this is intended to solve.

xsebek:
@jtdaugherty I updated the description. 👍 Simply put I run into a `fromJust` error and I wished there was more information. All I needed is in the rest of the printed line, so I made the code safely handle the situation.

jtdaugherty:
Oh, so the higher-level problem is that you had a key event in existence that was not in the `KeyEvents` mapping so it had no name, and the pretty-printer didn't handle that well.

jtdaugherty:
Thanks for your patch!

jtdaugherty:
This is now released in 2.3.2.

PR #515, opened 2024-09-28
Added instance of Splittable to default List type

isaac025:
Added the instance of `Splittable` (the same way it is defined over at Prelude) for the default list type of Haskell `[]`. 

I added it because I do not want to use Vector or Sequence (necessarily), however, if this is something you'd like to avoid, please feel free to ignore this PR. In that case, why is it that you do not want an instance of `Splittable` for `[]`? If you do not mind sharing :)

Thanks, and I hope this is helpful!

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -168,6 +168,10 @@ instance Splittable V.Vector where
 instance Splittable Seq.Seq where
     splitAt = Seq.splitAt
 
+-- /O(n)/ 'splitAt
+instance Splittable [] were
+    splitAt n xs = (take n xs, drop n xs)
+
 -- | Ordered container types where the order of elements can be
 -- reversed. Only required if you want to use 'listReverse'.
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Hi @isaac025 - thanks for writing this up!

The main reason there isn't already a `Splittable` instance for `[]` is because the performance of `splitAt` would be `O(n)`, which is considerably worse than that of `Vector` and `Seq`. In addition, providing an instance in the library communicates that it's okay to use `[a]` for the contents of `List`, which I want to dissuade so that bad performance doesn't result in a bad user experience.

I think I'd still rather keep things that way, which means I can think of two options in your case:
* Create orphan instances of `Splittable` etc. in your codebase and use `[]`, or
* Convert your `[]` to a `Vector` at `List` construction time. (You don't need to actually use `Vector` anywhere else; you could just convert at the right spot, although that would have a performance cost so it'd be good to think about that.)

I hope this helps!

PR #522, opened 2024-12-06
Support ghc 9.12

erikd:
Needed a `cabal.project` file containing the following to make it compile:

```
packages:
  .

tests: True

if impl (ghc >= 9.12)
  allow-newer:
    , hashable:base
    , integer-logarithms:base
    , integer-logarithms:ghc-prim
    , microlens-th:template-haskell
    , scientific:base
    , scientific:template-haskell
    , splitmix:base
    , time:template-haskell
    , unordered-containers:hashable
    , unordered-containers:template-haskell
```


erikd:
> Thank you! Any idea why the first CI build failed?

I think its transient. I have touch the top commit and force pushed.

erikd:
Damn, failed again. Its fails to find the required version of ghc (9.12.0.20241128) but I have updated a bunch of others like that and then were fine.

jtdaugherty:
There are two entries that build with the same version of GHC (9.10.0.20241128), but one is failing and one isn't (or at least it didn't on the initial build). Do you know the difference between them? I am inclined to remove the one that's failing if the other one continues to work.

erikd:
I have no idea why there are two. :confused:

jtdaugherty:
Okay, thanks for this patch. I'll keep an eye on it and see about removing one if it's consistently failing for environment reasons.

erikd:
Thanks @jtdaugherty . Would appreciate a Hackage metadata edit as well, unless you plan a new release in the near future.

jtdaugherty:
@erikd I made a metadata update for 2.6.

PR #526, opened 2025-03-06
Added str import in case Haddock couldn't the str file, moved simpleA…

liamzee:
…pp closer to the App datatype for greater visibility, increased explicitness on simpleApp's description.

Actually, I recently hacked together a primitive Tetris game with Brick, and one of the issues I disliked was how difficult it was to build a simple full-fledged program because I couldn't find a good defaultApp configuration which I could override.

Given the choice, actually, I'd rather take away the widget parameter from simpleApp, defaulting to "str \"Hello, Brick!\"", but to avoid others having similar issues, I moved simpleApp next to the App datatype, added str imports so Haddock could point to the widget.

### DIFF
--- a/cabal.project.local
+++ b/cabal.project.local
@@ -1,0 +1,1 @@
+tests: True
--- a/cabal.project.local
+++ b/cabal.project.local
@@ -1,1 +1,1 @@
-tests: True
+
--- a/src/Brick/Main.hs
+++ b/src/Brick/Main.hs
@@ -2,13 +2,13 @@
 {-# LANGUAGE RankNTypes #-}
 module Brick.Main
   ( App(..)
+  , simpleApp
   , defaultMain
   , customMain
   , customMainWithVty
   , customMainWithDefaultVty
   , simpleMain
   , resizeOrQuit
-  , simpleApp
 
   -- * Event handler functions
   , continueWithoutRedraw
@@ -85,6 +85,7 @@ import Brick.BChan (BChan, newBChan, readBChan, readBChan2, writeBChan)
 import Brick.Types.EventM
 import Brick.Types.Internal
 import Brick.Widgets.Internal
+import Brick.Widgets.Core (str)
 import Brick.AttrMap
 
 -- | The library application abstraction. Your application's operations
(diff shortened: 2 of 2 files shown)


jtdaugherty:
Hi @liamzee - thanks for submitting this.

I'm going to decline to merge this on a few grounds:
* Aside from the `cabal.project.local` fix, the commit here makes two unrelated changes that should be in one commit each. (The first changes the export ordering and the second adds the `str` import and changes some documentation.) A rule of thumb is that when you find yourself talking about multiple changes in the commit message, that's a good sign that those changes should be committed separately. The easiest way to do that is to use `git add -p` rather than `git add`, for example. Interactive commit interfaces typically also offer the ability to select specific hunks when staging a commit. I don't want anyone to have to rewrite their commits, so I'm mentioning this in case you make future contributions. Each commit needs to contain only the hunks related to that logical change.
* The documentation change isn't consistent with the style I'd like to use, although I want to acknowledge you might have found it helpful if it was there. I would rather focus on making sure that the other documentation I've incorporated (the User Guide especially) provides enough to help people get started comfortably. If you have ideas on how the User Guide could have been more helpful in your case, I'd love to discuss that!

liamzee:
TBH a Quick Start guide might have been useful. Brick is actually small enough that branching out from a quickstart is enough.

Some reexports from Vty might also have been helpful, in order to reduce the space of the library that has to be learned (i.e, it should be acceptable to know Brick, but not Vty).

jtdaugherty:
> TBH a Quick Start guide might have been useful.

Thanks, I'll give this some thought!

> Some reexports from Vty might also have been helpful, in order to reduce the space of the library that has to be learned (i.e, it should be acceptable to know Brick, but not Vty).

I'm afraid that it's just the case that knowing at least some of the Vty is essential for using Brick. That's made clear in the docs right at the start. While I understand that re-exports might help with discovering which bits of the Vty are needed, I prefer to avoid re-exports because they make it harder to change the API of the re-exporting library when the underlying library changes. At the end of the day, there is just a lot to learn about Vty and Brick, and I'm hoping that the extensive Haddock docs, User Guide, and dozens of demo programs I've created will be enough.

PR #530, opened 2025-06-21
doc: clarify that Vty from customMainWithDefaultVty needs shutdown

frasertweedale:
I wondered whether perhaps `customMainWithDefaultVty` ought to shut down the vty it creates, but since the return type includes the Vty, we probably should leave it as is, and just update the doc to clarify that it needs to be shut down.

### DIFF
--- a/src/Brick/Main.hs
+++ b/src/Brick/Main.hs
@@ -238,6 +238,10 @@ customMain initialVty buildVty mUserChan app initialAppState = do
 
 -- | Like 'customMainWithVty', except that Vty is initialized with the
 -- default configuration.
+--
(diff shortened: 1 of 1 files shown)


jtdaugherty:
Yes, the `shutdown` call is deferred to the caller specifically because otherwise the returned `Vty` handle would not be useful. The use case for those functions is one in which the caller wants to maintain control of the terminal after the Brick application halts but before other work is carried out by the caller that would require control of the terminal (and in which a `shutdown`-induced state restoration would be visually undesirable).

As for the change here, my hope had been that the docs imply this in the chain of "Like <other function>" but I'm happy to merge this to make it clearer. Thanks!

PR #536, opened 2025-09-20
Regenerate CI config

jhrcek:
It seems like github actions CI hasn't been working for some time.
This is likely because the generated config uses ubuntu-20 images, which have been recently decomissioned (see https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/). Regenerating config with more recent version of haskell-ci should fix this (uses ubuntu 24 by default).


brick.cabal
```diff
@@ -59,7 +59,7 @@ extra-doc-files:     README.md,
 
 Source-Repository head
   type:     git
-  location: git://github.com/jtdaugherty/brick.git
+  location: http://github.com/jtdaugherty/brick
```
jhrcek, line 62: Fixing `cabal check` warning (executed as part of generated haskell-ci config)
```
Error: [git-protocol] Cloning over git:// might lead to an arbitrary code
execution vulnerability. Furthermore, popular forges like GitHub do not
support it. Use https:// or ssh:// instead.
```

.github/workflows/haskell-ci.yml
```diff
@@ -214,14 +190,15 @@ jobs:
           touch cabal.project.local
           echo "packages: ${PKGDIR_brick}" >> cabal.project
           echo "package brick" >> cabal.project
-          echo "    ghc-options: -Werror=missing-methods" >> cabal.project
+          echo "    ghc-options: -Werror=missing-methods -Werror=missing-fields" >> cabal.project
+          if [ $((HCNUMVER >= 90400)) -ne 0 ] ; then echo "package brick" >> cabal.project ; fi
+          if [ $((HCNUMVER >= 90400)) -ne 0 ] ; then echo "    ghc-options: -Werror=unused-packages" >> cabal.project ; fi
```
jhrcek, line 195: Recent version of haskell-ci adds this warning, which causes build error if there are unused dependencies in cabal file - for that reason I cleaned them up in cabal file to make fix build failures caused by this warning.

jtdaugherty:
Thank you for your work on this!

PR #537, opened 2025-09-30
Preliminary support for ghc-9.14

erikd:
Currently using ghc-9.14.0.20250908.

**Probably not worth merging this until after the first full release of ghc-9.14.**


erikd:
The final `ghc9.14.1` release is at least a month or two away to the best of my knowledge.

PR #550, opened 2026-08-09
Make `GenericList` allow wrapping from first to last element and viceversa

Aster89:
With this PR I have

  - introduced a `HandleListEventConfig` ADT that contains, for now, just a `Bool` to tell whether moving up/down from the first/last element of the list should wrap, i.e. move to the last/first of the entries,

  - rewritten `handleListEvent` in terms of a new `handleListEvent'` that accepts `HandleListEventConfig` as its first input and,

      - if the `hleWrap :: Bool` in the `HandleListEventConfig` is `False`, it does what the pre-existing `handleListEvent` so far,

      - if that is `True`, it enables wrapping for `KUp`/`KDown`,`KPageUp`/`KPageDown`.

This means no change for existing clients, but all clients now also have the opportunity to enable wrapping for list navigation if they want to.

https://github.com/haskell/ghcup-hs/pull/1379 is where I'd like to use this new feature.

### DIFF
--- a/src/Brick/Widgets/List.hs
+++ b/src/Brick/Widgets/List.hs
@@ -28,6 +28,7 @@ module Brick.Widgets.List
 
   -- * Handling events
   , handleListEvent
+  , setScrollWrap
   , handleListEventVi
 
   -- * Lenses
@@ -112,6 +113,12 @@ import Brick.AttrMap
 -- * Home/end keys: move cursor of selected item to beginning or end of
 --   list
 --
+-- Both up/down and page up/page down movements are subject to wrapping
+-- if the given list's 'listScrollWrap' is 'True', i.e. moving "below/above"
+-- the first/last element of the list selects the last/first element of
+-- the list; otherwise the movements will clamp at @0@ and at @length list - 1@,
+-- i.e.  at the indices of the first and last elements of the list.
+--
 -- The 'List' type synonym fixes @t@ to 'V.Vector' for compatibility
 -- with previous versions of this library.
 --
@@ -133,6 +140,9 @@ data GenericList n t e =
          -- ^ The list's name.
          , listItemHeight :: Int
          -- ^ The height of an individual item in the list.
+         , listScrollWrap :: Bool
+         -- ^ Whether moving beyond a list's first/last element
+         -- should wrap to the last/first element.
          } deriving (Functor, Foldable, Traversable, Show, Generic)
 
 suffixLenses ''GenericList
@@ -263,7 +273,8 @@ listSelectedAttr = listAttr <> attrName "selected"
 listSelectedFocusedAttr :: AttrName
 listSelectedFocusedAttr = listSelectedAttr <> attrName "focused"
 
--- | Construct a list in terms of container 't' with element type 'e'.
+-- | Construct a list, with wrapping initially disabled, in terms of
+-- container 't' with element type 'e'.
 list :: (Foldable t)
      => n
      -- ^ The list name (must be unique)
@@ -276,7 +287,7 @@ list :: (Foldable t)
 list name es h =
     let selIndex = if null es then Nothing else Just 0
         safeHeight = max 1 h
-    in List es selIndex name safeHeight
+    in List es selIndex name safeHeight False
 
 -- | Render a list using the specified item drawing function.
 --
@@ -465,8 +476,8 @@ listReplace es idx l =
             | otherwise = 0
     in l' & listSelectedL .~ newSel
 
--- | Move the list selected index up by one. (Moves the cursor up,
--- subtracts one from the index.)
+-- | Move the list selected index up by one (moves the cursor up,
+-- subtracts one from the index), subject to wrapping.
 listMoveUp :: (Foldable t, Splittable t)
            => GenericList n t e
            -> GenericList n t e
@@ -477,8 +488,8 @@ listMovePageUp :: (Foldable t, Splittable t, Ord n)
                => EventM n (GenericList n t e) ()
 listMovePageUp = listMoveByPages (-1::Double)
 
--- | Move the list selected index down by one. (Moves the cursor down,
--- adds one to the index.)
+-- | Move the list selected index down by one (moves the cursor down,
+-- adds one to the index), subject to wrapping.
 listMoveDown :: (Foldable t, Splittable t)
              => GenericList n t e
              -> GenericList n t e
@@ -489,7 +500,8 @@ listMovePageDown :: (Foldable t, Splittable t, Ord n)
                  => EventM n (GenericList n t e) ()
 listMovePageDown = listMoveByPages (1::Double)
 
--- | Move the list selected index by some (fractional) number of pages.
+-- | Move the list selected index by some (fractional) number of pages,
+-- subject to wrapping.
 listMoveByPages :: (Foldable t, Splittable t, Ord n, RealFrac m)
                 => m
                 -> EventM n (GenericList n t e) ()
@@ -506,8 +518,9 @@ listMoveByPages pages = do
 -- | Move the list selected index.
 --
 -- If the current selection is @Just x@, the selection is adjusted by
--- the specified amount. The value is clamped to the extents of the list
--- (i.e. the selection does not "wrap").
+-- the specified amount. Whether the value is clamped to the extents of the
+-- list or not (i.e. whether the selection doesn't wrap around the list or it
+-- does), is determined by the list itself (see 'GenericList' for more info).
 --
 -- If the current selection is @Nothing@ (i.e. there is no selection)
 -- and the direction is positive, set to @Just 0@ (first element),
@@ -528,7 +541,10 @@ listMoveBy amt l =
             Nothing
                 | amt > 0 -> 0
                 | otherwise -> length l - 1
-            Just i -> max 0 (amt + i)  -- don't be negative
+            Just i
+                | let wrap = l ^. listScrollWrapL
+                , wrap -> (amt + i) `mod` length l
+                | otherwise -> max 0 (amt + i)  -- don't be negative
     in listMoveTo target l
 
 -- | Set the selected index for a list to the specified index, subject
@@ -692,3 +708,8 @@ listModify :: (Traversable t, Splittable t, Semigroup (t e))
            -> GenericList n t e
            -> GenericList n t e
 listModify f = listSelectedElementL %~ f
+
+-- | Sets the 'listScrollWrap' flag to the given 'Bool' for the given
+-- @GenericList n t e@.
+setScrollWrap :: (Traversable t, Splittable t, Semigroup (t e)) => Bool -> GenericList n t e -> GenericList n t e
+setScrollWrap b = listScrollWrapL .~ b
--- a/tests/List.hs
+++ b/tests/List.hs
@@ -220,6 +220,20 @@ prop_moveUp ops l =
         len = length l''
     in maybe (len == 0) (== 0) (l'' ^. listSelectedL)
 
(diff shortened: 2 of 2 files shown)


jtdaugherty:
Hi, thanks for submitting this!

Thinking about this a bit, I think I would prefer to make this configurability a field in the list itself rather than something that is configured whenever the event handler is called.

That would entail something like this:

* Add a field `listScrollWrap` to the `GenericList` data type
* Modify the `list` constructor so that newly-created lists have `listScrollWrap` set to `False` to preserve existing behavior
* Add a transformation function `setScrollWrap :: Bool -> GenericList n t e -> GenericList n t e` that sets `listScrollWrap`
* Have `handleListEvent` use `listScrollWrap` to determine scrolling behavior

Looking at the diff, it occurs to me that there might be a bit more to think about here. The functions that `handleListEvent` calls are intended to be used outside of `handleListEvent` to trigger the same underlying behaviors -- for example, `listMoveUp` is the implementation for `KUp`, but using `listMoveUp` on its own should have the same behavior as handling a `KUp` event. But if scrolling can be configured to wrap, then that suggests that "moving up" or "moving down" outside the context of `handleListEvent` ought to honor the wrapping setting. That suggests to me that functions like `listMoveUp` ought to do the wrapping, not `handleListEvent`.

Would you be willing to rework this as described above, so that the API matches what I outlined and so that the wrapping behavior is implemented within the various functions called by `handleListEvent` rather than in `handleListEvent` itself?

In addition, there are tests in `tests/List.hs` that need to be updated to check that the wrapping configuration is honored.

One last request: if you do choose to rewrite this, please do so without using `ViewPatterns`.

Aster89:
@jtdaugherty, you meant something like this?

You said
> That suggests to me that functions like `listMoveUp` ought to do the wrapping, not `handleListEvent`.

but then why not `listMoveBy`? That's what I've done.

Let me know what you think. When you gimme green, I'll update the doc and see what to do with tests.

src/Brick/Widgets/List.hs
```diff
@@ -196,7 +200,7 @@ instance Reversible Seq.Seq where
 handleListEvent :: (Foldable t, Splittable t, Ord n)
                 => Event
                 -> EventM n (GenericList n t e) ()
-handleListEvent e =
+handleListEvent e = do
```
jtdaugherty: Looks like this might be left over from a previous change.

src/Brick/Widgets/List.hs
```diff
@@ -692,3 +698,6 @@ listModify :: (Traversable t, Splittable t, Semigroup (t e))
            -> GenericList n t e
            -> GenericList n t e
 listModify f = listSelectedElementL %~ f
+
+setScrollWrap :: (Traversable t, Splittable t, Semigroup (t e)) => Bool -> GenericList n t e -> GenericList n t e
+setScrollWrap b = listScrollWrapL %~ const b
```
jtdaugherty: FWIW this is equivalent to `listScrollWrapL .~ b`.

jtdaugherty:
> but then why not listMoveBy? That's what I've done.

Looking at the latest diff, it makes sense to me that `listMoveBy` would be wrapping-aware since it's the workhorse for many of the other functions (I had forgotten about that when I mentioned the more specific ones like `listMoveUp`!), although I think that behavior needs to depend on whether wrapping is enabled -- in the diff I'm seeing, wrapping behavior is always present regardless of the setting. If wrapping is not enabled, that function should behave as it did before, since we'd need to preserve all of the existing behaviors. Does that make sense? So that would entail restoring the removed Haddock comment but extending it to indicate how wrapping changes its behavior.

As far as other documentation changes when you get to them, I'd also love to see the functions that are implemented in terms of `listMoveBy` have their docstrings updated to lightly mention the implications of wrapping, such as by adding "(previous text), subject to wrapping." I'd also love to see a documentation update for the `List` data type that explains the notion of wrapping and how it can be configured.

Thanks again for working on this!

jtdaugherty (changes requested):

src/Brick/Widgets/List.hs
Aster89: Yep, sorry, I'll clean this up.

src/Brick/Widgets/List.hs
Aster89: Thanks, my knowledge of lenses, let alone operators (for remembering them, one necessarily needs a hashmap in the brain constructed with habit), is very scarce.

Aster89:
> in the diff I'm seeing, wrapping behavior is always present regardless of the setting.

I don't think I follow. The field I added, `listScrollWrap`, is set to `False` by the `list` constructor, as you said, and the new behavior in `listMoveBy`'s implementation is guarded by `| wrap`; if that's false, the old behavior is retained; `setScrollWrap` is the only way to set `listScrollWrap`, and that's exported, but using it or not is up to the caller.

After all, for making use of the new feature in GHCup, I have to change
```haskell
do modify $ L.setScrollWrap True
   L.handleListEvent (Vty.EvKey Vty.KUp [])
```
to this
```haskell
do modify $ L.setScrollWrap True
   L.handleListEvent (Vty.EvKey Vty.KUp [])
```
If I don't, there won't be any wrapping behavior.

Or am I not understanding the very code I've written, as it's late here :D

jtdaugherty:
> I don't think I follow. The field I added, listScrollWrap, is set to False by the list constructor, as you said, and the new behavior in listMoveBy's implementation is guarded by | wrap; if that's false, the old behavior is retained; setScrollWrap is the only way to set listScrollWrap, and that's exported, but using it or not is up to the caller.

Thanks, I missed this in the diff since I guess I was expecting some modified/red lines but that change only added new ones! I'll want to merge those `Just` cases after we're done for increased clarity, but it's okay with me to leave them as-is for now.

jtdaugherty:
```haskell
do modify $ L.setScrollWrap True
   L.handleListEvent (Vty.EvKey Vty.KUp [])
```
FWIW, while this will definitely work, it's probably best to move the `setScrollWrap` call to the spot where the list is created. At any rate, it looks like it was a trivial change to get `ghcup` using it, which is nice!

Aster89:
> I'll want to merge those `Just` cases after we're done for increased clarity

You mean like this?
```haskell
 Just i
     | let wrap = l ^. listScrollWrapL
     , wrap -> (amt + i) `mod` length l
     | otherwise -> max 0 (amt + i)  -- don't be negative
```
instead of
```haskell
Just i
    | let wrap = l ^. listScrollWrapL
    , wrap -> (amt + i) `mod` length l
Just i -> max 0 (amt + i)  -- don't be negative
```
Personally, I find the latter more readable, thanks to the horizontal spacing that gives me a quicker indication of different cases. But if you prefer the former, just confirm, and I'll make the change.

Aster89:
> ```haskell
> do modify $ L.setScrollWrap True
>    L.handleListEvent (Vty.EvKey Vty.KUp [])
> ```
> 
> FWIW, while this will definitely work, it's probably best to move the `setScrollWrap` call to the spot where the list is created. At any rate, it looks like it was a trivial change to get `ghcup` using it, which is nice!

I suppose you're right, but I don't know yet where the list is created, ahah. I need some time to figure that out :stuck_out_tongue_closed_eyes:

jtdaugherty:
> You mean like this?

Yes, because the `Just i` is common to both cases, and then the use of `otherwise` makes the "if-then-else" behavior of that whole construct a little easier to see. With the latter case, the spacing might look nicer, but the control flow is distributed over two case arms. So if you are willing to refactor it to your first variant, I would be grateful. Thank you!

src/Brick/Widgets/List.hs
```diff
@@ -1,10 +1,11 @@
...
+{-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveTraversable #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE FlexibleInstances #-}
-{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
```
jtdaugherty: This is a minor thing for future PRs: this change reordered the extension list, which makes it difficult to tell what happened. Probably this was because of your editor trying to be helpful. In the future, if this was an intended cosmetic change, it should be in a separate commit; if made when actually enabling or removing an extension, the list should not be re-ordered. That way, it makes it easier to tell in the review that this was cosmetic and functionality-preserving vs. a functional change.

src/Brick/Widgets/List.hs
```diff
@@ -263,7 +274,8 @@ listSelectedAttr = listAttr <> attrName "selected"
 listSelectedFocusedAttr :: AttrName
 listSelectedFocusedAttr = listSelectedAttr <> attrName "focused"
 
--- | Construct a list in terms of container 't' with element type 'e'.
+-- | Construct a non-wrapping list in terms of container 't' with element
+-- type 'e'.
```
jtdaugherty: I think it would be a little more accurate to say that this constructs a list with wrapping "initially disabled"; as written, I'm concerned it will be easy to interpret this to mean that the list it builds cannot be made to wrap later on.

src/Brick/Widgets/List.hs
```diff
@@ -528,7 +542,10 @@ listMoveBy amt l =
             Nothing
                 | amt > 0 -> 0
                 | otherwise -> length l - 1
-            Just i -> max 0 (amt + i)  -- don't be negative
+            Just i
+                | let wrap = l ^. listScrollWrapL
+                , wrap -> (amt + i) `mod` length l
+                | otherwise -> max 0 (amt + i)  -- don't be negative
```
jtdaugherty, line 547: Thanks for this change!

tests/Main.hs
```diff
@@ -2,9 +2,7 @@
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TypeFamilies #-}
 
-import Control.Applicative
 import Data.Bool (bool)
-import Data.Traversable (sequenceA)
```
jtdaugherty, line 7: Can you please remove these two hunks from the PR? These might generate warnings when present on some GHCs but cause build failures when absent on others. We can always make these conditional so they don't generate warnings on GHCs where they are redundant imports. (And if we did, they'd go in their own commit.)

jtdaugherty (changes requested):

tests/Main.hs
jtdaugherty: In fact, this is actually what's causing the build failure on at least a couple of the versions being tested in CI!

src/Brick/Widgets/List.hs
```diff
@@ -5,6 +5,7 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE MultiWayIf #-}
```
jtdaugherty: Is this needed? I don't see any `if |` in use in this commit.

Aster89: Yep, I _was_ using it, then forgot to remove. Removed now.

jtdaugherty: Gotcha! I figured that might be the case :)

jtdaugherty:
Merged - thank you for your work on this and for addressing my feedback!

jtdaugherty:
This change and an update to `ListDemo` are now released in 2.13!

PR #551, opened 2026-08-11
Sort list of extensions + (conditionally) remove unneeded imports

Aster89:


jtdaugherty:
Thanks for submitting this separately. A couple things I want to mention:

* I am willing to merge the extension reordering change, but in general I am less interested in that kind of code cleanup so I might decline such changes in the future.
* It looks like the bound used in the check might need to change, based on the CI build failures. Can you investigate?

jtdaugherty:
Also, I want to mention that it helps to make reviews easier (and history easier to work with) if each commit makes a single isolated change. In this case, reordering the extensions is a totally separate change than the conditional compilation fix, so they belong in separate commits. That makes it easier to treat them like separate changes from a review standpoint, and it makes it easier to selectively revert one if it causes trouble without reverting the other one, too!

Aster89:
> if each commit makes a single isolated change

To clarify, you only talking of _1 change per commit_, not _1 commit per PR_, right? As in, in this case, is it ok that this PR is for _2_ changes, given they're both easy-to-describe cleanups?

jtdaugherty:
> To clarify, you only talking of 1 change per commit, not 1 commit per PR, right? As in, in this case, is it ok that this PR is for 2 changes, given they're both easy-to-describe cleanups?

Ah yes, that's what I meant: one (isolated/logical) change per commit, but a PR might well need to have lots of commits, not all of which are necessarily related (although it helps if they are, for merging purposes). And yes, I'm okay with this PR having these two changes. Thanks for checking!

